PR #1, opened 2023-09-20
Add more hotkeys, update readme/changleog

netniV:
This PR adds some README/CHANGELOG items and attempts to add a few more hotkeys for useful daily tasks. 

**NOTE:** *Some items are still work in progress as they do not have a SectionID or require extra implementation that I'm not yet sure of.  These may ultimately be removed or require assistance in completing*

### DIFF
--- a/src/patches/parts/hotkeys.cc
+++ b/src/patches/parts/hotkeys.cc
@@ -172,9 +172,21 @@ void     ScreenManager_Update_Hook(auto original, ScreenManager* _this)
         }
       }
     } else if (GetKeyDownInt(KeyCode::G) && !is_input_focused()) {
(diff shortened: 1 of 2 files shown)


tashcan:
Hey, thanks for this.

I have just updated the SectionID enum with the latest version, that should give some of the missing ones. Currently still looking into what needs to be done to fix some of the screens not loading when doing a section change without first manually opening the section once.

netniV:
I suspect, based on what i've seen so far, the reason it works after the first time, is that the context or data args have been passed in once, so it uses the cached versions of them.  It may be that to display those sections, we need to make sure we have the context, but that could be tricky I guess.

I have now updated the PR (force a push to refresh it all). Also added the ability to hit escape to hide scanned items (thanks for the help on that).

PR #8, opened 2023-10-08
Customisable HotKeys

netniV:
.. and a whole bunch of other functionality like:

- mappable keys
- disable ship relocation when viewers are present
- allow manage ships
- renamed `_settings_parsed.toml` to `_runtime.vars` to prevent peoples confusion
- and more...


src/version.h
```diff
@@ -4,15 +4,16 @@
 #define VERSION_MAJOR               0
 #define VERSION_MINOR               5
 #define VERSION_REVISION            3
+#define VERSION_PATCH               2
```
tashcan, line 7: We don't do patch, as revision == patch

src/patches/key.cc
```diff
@@ -0,0 +1,278 @@
...
+{
+  return Key::Pressed(KeyCode::LeftAlt) || Key::Pressed(KeyCode::RightShift);
+}
+
+bool Key::HasCtrl()
+{
+  return Key::Pressed(KeyCode::LeftControl) || Key::Pressed(KeyCode::RightControl);
+}
+
+void Key::ClearInput()
```
tashcan, line 263: This doesn't really clear, this rather just removes the focus of the input element on the UI tree. This is a workaround for the game not doing that when using keyboard navigation using Escape.
I would suggest renaming this to `ClearInputFocus` or similar to make it clear what the the intent is.

src/patches/key.cc
```diff
@@ -0,0 +1,278 @@
...
+};
+
+Key::Key()
+{
+  // Map();
+}
+
+KeyCode Key::Parse(std::string_view key)
+{
+  auto wantedKey = absl::AsciiStrToUpper(key);
```
tashcan, line 185: This appears unused?

src/patches/parts/hotkeys.cc
```diff
@@ -150,105 +120,106 @@ void ScreenManager_Update_Hook(auto original, ScreenManager* _this)
         } else {
           fleet_bar->RequestSelect(ship_select_request);
         }
+        return;
       }
     }
   }
 
-  if (GetKeyDownInt(KeyCode::Escape)) {
+  if (Key::Pressed(KeyCode::Escape) && (Key::IsInputFocused || Hub::IsInChat)) {
```
tashcan, line 128: I am going to assume you intended to call the functions here instead of having a "always evaluates to true" as you are now checking the pointers of the function

src/patches/mapkey.h
```diff
@@ -0,0 +1,33 @@
...
+#include <map>
+#include <modifierkey.h>
+#include <patches/gamefunctions.h>
+#include <prime/KeyCode.h>
+#include <string>
+
+class MapKey
+{
+private:
+  static std::map<GameFunction, MapKey*> mappedKeys;
```
tashcan, line 14: As this is just an enum. It would be better to just have a std::array here instead of using a map.
Similarly, Is there a particular reason that we are storing a raw pointer here? Looks to me like you could just store the object directly, storing a pointer will add an additional indirection, while minor in perf, still adds up eventually.

_Side note: Generally you don't need map an can use/should prefer using unordered_map_

src/patches/key.cc
```diff
@@ -0,0 +1,278 @@
...
+      return (KeyCode)keycode;
+    }
+  }
+
+  return KeyCode::None;
+}
+
+bool Key::IsModifier(KeyCode key)
+{
+  return (std::find(Key::modifiers.begin(), Key::modifiers.end(), key) != Key::modifiers.end());
```
tashcan, line 197: I feel like having a vector here where we do a linear search every time to check the key is sub-optimal, I would probably change this to a simple if, as in
```c++
return key == KeyCode::LeftAlt || key == KeyCode::RightAlt ...
```

src/patches/key.cc
```diff
@@ -0,0 +1,278 @@
...
+}
+
+bool Key::IsModified()
+{
+  for (const auto modifier: Key::modifiers) {
```
tashcan, line 202: I get the feeling that we should probably cache modified on a per-frame basis. Doing repeated calls into the engine for this is most likely a performance cliff.

src/patches/mapkey.cc
```diff
@@ -0,0 +1,143 @@
...
+  this->Key = KeyCode::None;
+}
+
+MapKey* MapKey::Parse(std::string_view key)
+{
+  auto strippedKey = absl::StripTrailingAsciiWhitespace(key);
+  auto lowerKey    = absl::AsciiStrToUpper(strippedKey);
+  auto wantedKeys  = absl::StrSplit(lowerKey, "-", absl::SkipWhitespace());
+
+  auto mapKey = new MapKey();
```
tashcan, line 27: I would personally prefer this to just return a std::optional instead of a raw pointer that is never deleted. If that's not desired, at least a unique_ptr :)

src/patches/mapkey.cc
```diff
@@ -0,0 +1,143 @@
...
+
+  return output;
+}
+
+std::map<GameFunction, MapKey*> MapKey::mappedKeys = {};
+
+std::vector<std::string>  Shortcuts = {};
+std::vector<ModifierKey*> Modifiers = {};
+
+KeyCode Key;
```
tashcan, line 143: I am not quite sure what these dare doing here?

src/patches/modifierkey.cc
```diff
@@ -0,0 +1,101 @@
...
+
+  if (!this->Contains(modifier2)) {
+    this->Modifiers.emplace_back(modifier2);
+    this->Shortcuts.emplace_back(shortcut);
+  }
+}
+
+bool ModifierKey::IsPressed()
+{
+  for (auto modifier : this->Modifiers) {
```
tashcan, line 70: Similar to another comment. We should really think about caching key states per-frame at this point before we proceed with this, instead of doing an engine call every time we want to know the state. I suspect their implementation is sub-par at best.

src/patches/parts/hotkeys.cc
```diff
@@ -257,90 +228,76 @@ void ScreenManager_Update_Hook(auto original, ScreenManager* _this)
...
-  //  }
-  //}
-
-  // Config::Get().Load();
-
-  original(_this);
 }
 
-template <typename T> inline bool DidHideViewersOfType()
+template <typename T> inline bool CanHideViewersOfType(bool shouldHide = false)
```
tashcan, line 300: Not a huge fan of changing function behavior entirely based on a boolean param. Makes it not super clear on the callsite what is actually happening in most situations.
Especially when a 'Can' becomes a 'Do'

src/patches/mapkey.cc
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  if (mapKey->Key != KeyCode::None) {
+    return mapKey;
+  }
+
+  return nullptr;
+}
+
+bool MapKey::HasGameFunction(GameFunction gameFunction)
```
tashcan, line 54: As per my comment on `mapkey.h` a lot of these will have to change

src/patches/modifierkey.cc
```diff
@@ -0,0 +1,101 @@
...
+
+ModifierKey::ModifierKey() {}
+
+ModifierKey* ModifierKey::Parse(std::string_view key)
+{
+  auto strippedKey = absl::StripTrailingAsciiWhitespace(key);
+  auto upperKey    = absl::AsciiStrToUpper(strippedKey);
+  auto wantedKeys  = absl::StrSplit(upperKey, "-", absl::SkipWhitespace());
+
+  auto modifierKey = new ModifierKey();
```
tashcan, line 20: Same as the other thing. I fail to see a good reason for this to be a heap allocated object per modified key that is just a pointer that will never get freed

src/patches/parts/hotkeys.cc
```diff
@@ -515,7 +502,7 @@ void ChatMessageListLocalViewController_AboutToShow_Hook(ChatMessageListLocalVie
 
 void InitializeActions_Hook(auto original, void* _this)
 {
-  if (Config::Get().use_scopely_hotkeys && Config::Get().hotkeys_enabled) {
```
tashcan, line 518: What is the reasoning for removing this?
This is intended to disable hotkeys entirely, both the regular in-game ones plus ours.

KEYMAPPING.md
```diff
@@ -0,0 +1,84 @@
+# Star Trek Fleet Command - Community Patch
```
tashcan, line 1: This header doesn't need to be in this file :)

README.md
```diff
@@ -77,39 +82,65 @@ X | ExoComp
...
+## Installation / Runtme
+
+Both the DLL and the toml file must be placed in the STFC game folder which is always located
+at `C:\Games\Star Trek Fleet Command\Star Trek Fleet Command\default\game` as the following
+names:
+
+- community_patch_settings.toml
+- version.dll
+
+The pre-compiled DLL can be downloaded from the official assets hosted at:
```
tashcan, line 98: I would point to the Releases section here instead, as that is what will be used in addition to the eventual installer. Just a note to check the releases section is sufficient

src/patches/key.h
```diff
@@ -0,0 +1,32 @@
...
+#include <map>
+#include <prime/KeyCode.h>
+#include <string>
+#include <vector>
+
+class Key
+{
+private:
+  static const std::map<std::string, KeyCode> mappedKeys;
+  static const std::vector<KeyCode>           modifiers;
```
tashcan, line 13: We could probably get away with mapping this to a bit mask, that way we wouldn't have to do a linear scan to find the modifier.
Of course, that will require the other work I've mentioned to be in-place with regards to caching key-states per-frame somewhere.

tashcan: As an additional note. This is entirely static class. I would prefer to use a namespace instead for that.

src/patches/parts/hotkeys.cc
```diff
@@ -396,20 +394,20 @@ void ExecuteSpaceAction(FleetBarViewController* fleet_bar, bool (*GetKeyDownInt)
       if (pre_scan_widget
           && (pre_scan_widget->_visibilityController->_state == VisibilityState::Visible
               || pre_scan_widget->_visibilityController->_state == VisibilityState::Show)) {
-        did_pre_scan = true;
```
tashcan, line 399: There was a reason for this at some point, I forget what it was though

tashcan (changes requested):
I have left a few comments.

And maybe we should split out the contributing.md and readme.md update out of this which seems to be basically adding configurable hotkeys. (Then adjust PR title accordingly)

src/version.h
netniV: Patch is so I can track what beta testers have been using, this can always be 0 for release, if that's OK?

KEYMAPPING.md
netniV: There should always be a level 1 header for Markdown files.  I kept the same title for consistency.

src/patches/key.cc
netniV: Agreed.

src/patches/key.cc
netniV: Fixed.

src/patches/parts/hotkeys.cc
netniV: Fixed, thanks.

src/patches/parts/hotkeys.cc
netniV: Someone mentioned something in the chat, not entirely sure what is was now..... I can restore it as I don't think it makes a difference to what I've done.

src/patches/key.h
netniV: Not sure what you mean on the namespace front, but I have now removed the modifiers in favour of static code.

src/patches/parts/hotkeys.cc
netniV: Possibly because you were not returning on each action, which the new code does

src/patches/mapkey.cc
netniV: Not sure I follow this, but that may be down to my limited experience with this style of c++

src/patches/mapkey.cc
netniV: Not quite sure I follow what you're missing, they are the fields used by the class?  Shortcuts and Modifiers were being used for logging purposes originally, though I guess these could now become private fields.

src/patches/modifierkey.cc
netniV: This has now been revamped.  There is a new cached state for overall and individual results so we only call it once.

src/patches/parts/hotkeys.cc
netniV: I split them out into Can/DidHideViewer functions.  I guess we aren't likely to change the logic inside, it just duplicates that part of the code... unless we template it further and pass in an anonymous function but that seems overkill.

src/patches/mapkey.cc
netniV: Revamped where possible.  There's now a single gamefunction array rather than a map.

README.md
netniV: We should probably follow that up on the #6 PR.

tashcan:
Merged manuall

PR #20, opened 2024-02-19
Fix a couple of typos in README.md

gmn42:
I ran into these when I was following the directions to set up a build environment for the mod


tashcan:
Thanks.

PR #21, opened 2024-02-19
fix mklink syntax

gmn42:


tashcan:
Looks like this incorrectly adds abseil-cpp as a submodule which results in CI failing.

gmn42:
Oops, no idea how that got added, I missed it in the commit.  I think it may have happened when i switched vscode from main to dev.

PR #24, opened 2024-03-15
Move from CMake to XMake

tashcan:
I'm sick of dealing with CMake, especially as work on a possible mac port is kind of something I'd like to do.

This sets things up a little different compared to before


tashcan:
@netniV :)

netniV:
Oooo

PR #31, opened 2024-04-30
Allow hitting spacebar to cancel spooling up warp.

juhunter:


netniV (approved):

netniV:
@juhunter Can you please see #32 as this error caused me issue.  Seems we can't leave the call back parameter as nullptr.

PR #38, opened 2024-07-17
stfc path in readme adapted

Hofei90:
stfc path adapted


netniV:
So, the path in the README is correct for most default installations.  However, it CAN change which seems to happen if it finds an existing path or something.

PR #41, opened 2024-08-14
Add X-Instance-ID header

lightbull-stfc:
Due to the recent switch to generic/regional URLs, an instance-specific header is now required


netniV (approved):

netniV (approved):

mods/src/patches/parts/sync.cc
```diff
@@ -265,6 +266,7 @@ static std::wstring get_data_data(std::wstring session, std::wstring url, std::w
     list = sync_slist_append(CURL_TYPE_DOWNLOAD, list, "X-Api-Key", PRIME_API_KEY);
     list = sync_slist_append(CURL_TYPE_DOWNLOAD, list, "X-Unity-Version", UNITY_VERSION);
     list = sync_slist_append(CURL_TYPE_DOWNLOAD, list, "X-PRIME-VERSION", PRIME_VERSION);
+    list = sync_slist_append(CURL_TYPE_DOWNLOAD, list, "X-Instance-ID", std::format("{:03}", instanceId));
```
tashcan, line 269: std::format is not yet available on MacOS

netniV: is there an alternative to use?

PR #42, opened 2024-08-21
Allow passing a configuration filename [WIN-ONLY]

netniV:
This allow windows users to pass a configuration file vka te command line option "-ccm" and using the following value as a filename.

### DIFF
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -281,6 +281,20 @@ void parse_config_shortcut(toml::table config, toml::table& new_config, std::str
   spdlog::info("shortcut value {}.{} value: {}", section, item, shortcut);
 }
 
+#if _WIN32
+std::string ConvertWStringToString(const std::wstring& wstr)
+{
+  if (wstr.empty())
+    return std::string();
+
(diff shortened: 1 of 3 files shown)


mods/src/config.cc
```diff
@@ -274,15 +281,62 @@ void parse_config_shortcut(toml::table config, toml::table& new_config, std::str
...
+      for (int i = 0; i < argc - 1; ++i) {
+        if (std::wstring(argv[i]) == L"-ccm" && i + 1 < argc) {
+          // Found "-ccm", so take the next argument as the value
+          argValue = argv[i + 1];
+          break;
+        }
+      }
+
+      if (!argValue.empty()) {
+        configFile.assign(argValue.begin(), argValue.end());
```
tashcan, line 310: This is not doing proper conversions, files names should be UTF-8 encoded, this doesn't convert to UTF-8 and will result in garbage for non-ascii filenames

tashcan (changes requested):

mods/src/config.cc
netniV: thanks, I had a feeling that might be the case.

PR #44, opened 2024-09-14
Support for multiple sync targets

lightbull-stfc:
Add support for multiple sync targets. Legacy options `sync_url` and `sync_token` will be converted to `[sync.targets.default]`.

```YAML
[sync]

    [sync.targets.local]
    token = 'foo'
    url = 'http://localhost/sync/ingress/'

    [sync.targets.spocksclub]
    token = 'bar'
    url = 'https://spocks.club/sync/ingress/'
```

### DIFF
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -63,12 +63,6 @@ vsync = 1
 zoom = 2500
 
 [sync]
-# Token for auth on remote system
-token = ''
-
-# Url to post sync data to
-url = ''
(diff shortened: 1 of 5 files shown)


netniV:
This looks good from my perspective, but, lets keep the original sync paths for those who may have it configured and not realise that it's moved, add those as a primary target it not empty?  Maybe add a "DEPRECATED: " notice to the log file to say that these will be removed in a future version.

PR #74, opened 2025-08-12
wine docs and general revisions

gmn42:
Added info on wine/linux installation and other revisions to separate installation instructions for each platform and clarify config. Mac procedures should be verified.


gmn42:
sorry, messed this one up too

PR #77, opened 2025-08-24
Chat hotkey fixes

gmn42:
Two fixes for hotkeys in chat:
1) Adjust indexes for chat channels due to addition of Cadet chat in patch 78.  Allows chat channel selection hotkeys (^1, ^2, ^3) to work as expected.  (TODO: IsInChat() should also be updated to include Cadet chat, but this requires a SectionID enum definition in Hub.cc, and I'm not sure how to extract that.)
2) Remove ActionView keybind handling in chat.  It interferes with the emoji search function and does not seem to be needed any more (input box has been active when chat appears in all testing i've been able to do).  If retained, it should probably have a different keybind than ActionView, which is primarily for toggling target view modes, and code to disable it while emoji popup is active would be needed.

### DIFF
--- a/mods/src/patches/parts/hotkeys.cc
+++ b/mods/src/patches/parts/hotkeys.cc
@@ -344,14 +344,6 @@ void ScreenManager_Update_Hook(auto original, ScreenManager* _this)
         return chat_manager->OpenChannel(ChatChannelCategory::Private);
       }
     }
-
-    if (MapKey::IsDown(GameFunction::ActionView)) {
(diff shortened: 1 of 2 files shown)


gmn42:
Not sure why this isn't building on mac.  Error indicates abseil isn't getting linked correctly?  It worked prior to accepting the suggestion for the newbie chat, and I don't see any changes to the makefiles that would cause it.  Is it something in the new CI changes?

netniV:
Actually, I noticed that when you submitted the PR, it hadn't passed.  But I will look into why.

PR #111, opened 2026-03-13
Fix crashes with game client v48084

a5ehren:
Introduce SAFE_STATIC_DETOUR macro that validates IL2CPP method parameter counts before installing hooks, preventing crashes from game signature changes. Applied to all patch files replacing manual GetMethod/null-check patterns.

Key fixes:
- object_tracker: null guards for .ctor/OnDestroy resolution, GC finalizer null check, and signature scan size check before .get(0)
- sync: update ProcessResultInternal hook from 4 to 2 params (v48084 API change), disable on ARM64 where spud trampoline can't relocate prologue
- All patches now log actual parameter types on signature mismatch


netniV:
As there is a new PR, closing this one off.

PR #112, opened 2026-03-14
Add NULL checks to ObjectTracker to prevent crash on game updates

MBurchard:
## Summary

Game update v1.000.48084 crashes the mod at startup with SIGSEGV at address 0x0. The crash originates in `InstallObjectTrackers()`, the only patch that passes method pointers to `SPUD_STATIC_DETOUR` without validating them first. When `GetMethod()` returns nullptr (because the game binary changed), SPUD tries to read original instructions at address 0x0.

This PR adds defensive NULL checks following the same pattern used by all other 14 patches (isValidHelper + ErrorMsg):

- Validate class helper before accessing methods in `TrackObject<T>()`
- Check `GetMethod()` return values for `.ctor` and `OnDestroy` before hooking
- Guard `GC_register_finalizer_inner` call against nullptr
- Add bounds check on `find_in_module()` result before accessing match at index 0

With these checks, the mod logs missing methods as errors and continues instead of crashing. The ObjectTracker will not track objects when methods are missing, but all other patches remain functional.

## Test plan

- [x] `xmake build` succeeds on macOS (arm64)
- [x] Game starts without SIGSEGV, mod initialises and logs missing methods as errors
- [x] Verified on game version v1.000.48084

### DIFF
--- a/mods/src/patches/parts/object_tracker.cc
+++ b/mods/src/patches/parts/object_tracker.cc
@@ -1,5 +1,6 @@
 #include <il2cpp/il2cpp_helper.h>
 
+#include "errormsg.h"
 #include "prime/AllianceStarbaseObjectViewerWidget.h"
 #include "prime/AnimatedRewardsScreenViewController.h"
 #include "prime/ArmadaObjectViewerWidget.h"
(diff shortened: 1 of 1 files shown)


a5ehren:
this does the same thing as #111 without the other fixes

MBurchard:
Fair enough, #111 covers this and more. Closing in favour of yours.

PR #113, opened 2026-03-14
Write log file to OS-standard location instead of game directory

MBurchard:
## Summary

The log file was written relative to the game's working directory, making it hard to find when debugging user-reported issues. This PR writes the log to OS-standard locations:

- **macOS:** `~/Library/Logs/com.stfcmod.startrekpatch/community_patch.log`
- **Windows:** `%LOCALAPPDATA%\stfcmod\Logs\community_patch.log`

The log directory is created automatically on startup. The `-ccm` command line override continues to work as before.

## Changes

- `file.cc`: Platform-specific log path calculation in the default (non-override) case
- `file.h`: Add `ShlObj.h` include for `SHGetKnownFolderPath` on Windows

## Test plan

- [x] `xmake build` succeeds on macOS (arm64)
- [x] Log file appears at `~/Library/Logs/com.stfcmod.startrekpatch/community_patch.log`
- [x] Log directory is created if it does not exist
- [ ] Windows build and log path verification (not yet tested)

### DIFF
--- a/mods/src/file.cc
+++ b/mods/src/file.cc
@@ -192,7 +192,25 @@ void File::Init()
     if (File::override) {
       cacheNameLog = configPath.replace_extension(FILE_EXT_LOG).string();
     } else {
-      cacheNameLog = std::string(FILE_DEF_LOG);
+#if !_WIN32
+      const std::filesystem::path libraryPath =
+          fm::FolderManager::pathForDirectory(fm::NSLibraryDirectory, fm::NSUserDomainMask);
+      const auto log_dir = libraryPath / "Logs" / "com.stfcmod.startrekpatch";
+      std::error_code ec;
+      std::filesystem::create_directories(log_dir, ec);
+      cacheNameLog = (log_dir / FILE_DEF_LOG).string();
+#else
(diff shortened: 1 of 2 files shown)


a5ehren:
I'd probably prefer if this were MacOS only, the windows Mod just dumping in the same place as version.dll is fine.

MBurchard:
I disagree on principle. Log files shouldn't pollute the game directory on any platform. On Windows, %LOCALAPPDATA% is the standard location for application logs and avoids permission issues with Program Files paths.

That said, I don't want to hold up your work on #111. If you'd rather keep the Windows log path as-is, I'll close this PR and leave it to you.

a5ehren:
Oh I'm not gonna touch the logging code at all. I'll leave this one up to @netniV

netniV:
When using the -ccm option, the log files are placed in the same directory as the toml file.  This is our standard and makes life easier when requesting details from users who do not know what they are doing.  So, whilst %appdata% may be standard for data under windows, I'm not convinced on making that change just yet.

But please try to convince me why that's important to move, and I may change my mind.

PR #115, opened 2026-03-17
V48084 fixes

a5ehren:
Smaller version of my other PR, with a fix for the GC_finalizer

### DIFF
--- a/mods/src/patches/parts/sync.cc
+++ b/mods/src/patches/parts/sync.cc
@@ -34,6 +34,7 @@
 #include <atomic>
 #include <chrono>
 #include <condition_variable>
+#include <filesystem>
 #include <format>
 #include <fstream>
 #include <mutex>
@@ -666,7 +667,8 @@ static void load_previously_sent_logs()
   previously_sent_battlelogs.set_capacity(300);
 
   try {
-    std::ifstream file(File::Battles(), std::ios::in | std::ios::binary);
+    const std::filesystem::path path(File::MakePath(File::Battles()));
+    std::ifstream              file(path, std::ios::in | std::ios::binary);
     if (!file.is_open()) {
       spdlog::warn("Failed to open battles file (not found or not readable); starting with empty cache");
       return;
@@ -698,7 +700,8 @@ static void save_previously_sent_logs()
   }
 
   try {
-    std::ofstream file(File::Battles(), std::ios::out | std::ios::binary | std::ios::trunc);
+    const std::filesystem::path path(File::MakePath(File::Battles(), true));
+    std::ofstream               file(path, std::ios::out | std::ios::binary | std::ios::trunc);
     if (!file.is_open()) {
       spdlog::error("Failed to open battles file for writing");
       return;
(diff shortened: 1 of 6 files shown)


mods/src/patches/parts/sync.cc
```diff
@@ -2054,29 +2064,30 @@ void DataContainer_ParseRtcPayload(auto original, void* _this, bool incrementalJ
...
+                                                   ServiceResponse* service_response, const MethodInfo* method)
 {
   auto *const entity_groups = service_response->EntityGroups;
   for (int i = 0; i < entity_groups->Count; ++i) {
     auto *const entity_group = entity_groups->get_Item(i);
     HandleEntityGroup(entity_group);
   }
 
-  return original(_this, http_response, service_response, callback, callback_error);
+  return original(_this, parsing_context, service_response, method);
```
netniV, line 2077: So this would still seem to be wrong as the `ProcessResultInternal` only has two parameters:
```c#
   [Token(Token = "0x60004C1")]
    protected override void ProcessResultInternal(
      IParsingContext parsingContext,
      TServiceResponse serviceResponse)
    {
    }
```

mods/src/patches/parts/sync.cc
```diff
@@ -2054,29 +2064,30 @@ void DataContainer_ParseRtcPayload(auto original, void* _this, bool incrementalJ
...
+                                                       ServiceResponse* service_response, const MethodInfo* method)
 {
   auto *const entity_groups = service_response->EntityGroups;
   for (int i = 0; i < entity_groups->Count; ++i) {
     auto *const entity_group = entity_groups->get_Item(i);
     HandleEntityGroup(entity_group);
   }
 
-  return original(_this, service_response);
+  return original(_this, parsing_context, service_response, method);
```
netniV, line 2089: Same here...
```c#
   [Token(Token = "0x60004C8")]
    protected virtual void ParseBinaryObjectsHelper(
      IParsingContext parsingContext,
      TServiceResponse serviceResponse)
    {
    }
```

mods/src/patches/parts/sync.cc
```diff
@@ -2023,6 +2026,13 @@ void DataContainer_ParseEntitySlotsData(auto original, void* _this, EntityGroup*
   return original(_this, group);
 }
 
+void DataContainer_ParseSlotData(auto original, void* _this, void* entity_slot, Il2CppString* channel_id)
+{
+  // TODO: figure out what is in this entity_slot struct.
+  return original(_this, entity_slot, channel_id);
+}
+
+#if 0
```
netniV: Why are we completely removing the Parsing of RTC data?

mods/src/patches/parts/sync.cc
```diff
@@ -2249,16 +2261,18 @@ void InstallSyncPatches()
...
+      SPUD_STATIC_DETOUR(ptr, DataContainer_ParseSlotData);
     }
 
-    if (auto *const ptr = slot_data_container.GetMethod("ParseSlotRemovedJson"); ptr == nullptr) {
-      ErrorMsg::MissingMethod("SlotDataContainer", "ParseSlotRemovedJson");
+    if (auto *const ptr = slot_data_container.GetMethod("RemoveSlotData"); ptr == nullptr) {
+      ErrorMsg::MissingMethod("SlotDataContainer", "RemoveSlotData");
     } else {
-      SPUD_STATIC_DETOUR(ptr, DataContainer_ParseRtcPayload);
+      SPUD_STATIC_DETOUR(ptr, DataContainer_ParseSlotData);
```
netniV, line 2274: You are hooking RemoveSlotData but yet calling the same ParseSlotData method?

netniV (changes requested):

mods/src/patches/parts/sync.cc
a5ehren: The methods that used this hook no longer exist, so I have to either remove or ifdef it to be able to compile.

mods/src/patches/parts/sync.cc
a5ehren: I am just following the same pattern as the previous hooks. Used the same hook for Update/RemoveJson.

mods/src/patches/parts/sync.cc
a5ehren: apparently claude told itself that these were C# generics last time, dunno. I'll take it off.

mods/src/patches/parts/sync.cc
```diff
@@ -666,7 +667,8 @@ static void load_previously_sent_logs()
   previously_sent_battlelogs.set_capacity(300);
 
   try {
-    std::ifstream file(File::Battles(), std::ios::in | std::ios::binary);
+    const std::filesystem::path path(File::MakePath(File::Battles()));
+    std::ifstream              file(path, std::ios::in | std::ios::binary);
```
a5ehren, line 671: this is to fix the battlelog save/load on mac, btw

mods/src/patches/parts/sync.cc
lightbull-stfc: Both payloads from Update & Remove were passed into the same function since the disambiguation happens later in https://github.com/a5ehren/stfc-mod/blob/131a4255fd8814da83b8d75e000272170de42ebf/mods/src/patches/parts/sync.cc#L2045

mods/src/patches/parts/sync.cc
a5ehren: Apparently this is legit. Game crashes without it.

PR #124, opened 2026-04-12
feat: add double-tap Escape to exit when disable_escape_exit is enabled

Guffawaffle:
## Summary

Adds a double-tap Escape mechanism to exit the game even when `disable_escape_exit` is enabled.

Currently, when `disable_escape_exit = true`, all Escape key presses are unconditionally swallowed. This means users have no keyboard-only way to trigger the exit prompt — they must use the F10 hard kill (which calls `TerminateProcess` with no cleanup).

This PR adds a configurable double-tap window: press Escape twice within `escape_exit_timer` milliseconds to fall through to the original handler and trigger the native exit dialog.

## New Config Option

```toml
[ui]
# When disable_escape_exit is true, allow double-tap Escape to still trigger
# the exit prompt. Set the maximum time in milliseconds between presses.
# Set to 0 to disable double-tap (Escape is fully blocked).
# Has no effect when disable_escape_exit is false.
escape_exit_timer = 0
```

## Implementation

Uses the same `std::chrono::steady_clock` double-tap pattern already established for ship selection (`select_clock` in `hotkeys.cc` line 75). A static `escape_clock` tracks the last Escape press time — first press is swallowed, second press within the timer window falls through to `original(_this)`.

## Files Changed

| File | Change |
|------|--------|
| `mods/src/patches/parts/hotkeys.cc` | Double-tap timer logic in the `disable_escape_exit` guard |
| `mods/src/config.h` | Added `escape_exit_timer` field |
| `mods/src/config.cc` | Load `escape_exit_timer` from `[ui]` section |
| `mods/src/defaultconfig.h` | Default value: `0` (preserves existing behavior) |
| `example_community_patch_settings.toml` | Documented the new option |

## Behavior

| `disable_escape_exit` | `escape_exit_timer` | Effect |
|---|---|---|
| `false` | *(ignored)* | Escape works normally (game default) |
| `true` | `0` (default) | Escape fully blocked (existing behavior preserved) |
| `true` | `500` | Double-tap Escape within 500ms to exit |

## Testing

Built and tested on Windows with `releasedbg` mode. Confirmed:
- Single Escape press is swallowed when `disable_escape_exit = true`
- Double-tap within 500ms triggers the exit prompt
- Setting `escape_exit_timer = 0` fully blocks Escape (backward compatible)
- No impact on Escape behavior when `disable_escape_exit = false`

### DIFF
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -310,9 +310,13 @@ always_skip_reveal_sequence = true
 # Prevent "escape" from prompting to exit the game
 disable_escape_exit = true
 
-# When disable_escape_exit is true, allow double-tap Escape to still trigger
-# the exit prompt. Set the maximum time in milliseconds between presses.
-# Set to 0 to disable double-tap (Escape is fully blocked).
+# When disable_escape_exit is true, this lets you double-tap Escape to still
+# get the exit prompt. The value is the maximum gap in milliseconds between
+# two Escape presses to count as a double-tap.
+#
+#   0   = double-tap disabled, Escape is fully blocked (default)
+#   500 = press Escape twice within half a second to trigger the exit prompt
+#
 # Has no effect when disable_escape_exit is false.
 escape_exit_timer = 0
 
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -310,6 +310,12 @@ always_skip_reveal_sequence = true
 # Prevent "escape" from prompting to exit the game
 disable_escape_exit = true
 
+# When disable_escape_exit is true, allow double-tap Escape to still trigger
+# the exit prompt. Set the maximum time in milliseconds between presses.
+# Set to 0 to disable double-tap (Escape is fully blocked).
+# Has no effect when disable_escape_exit is false.
+escape_exit_timer = 0
+
 # ⚠️ EXPERIMENTAL: Prevent the first popup advert from appearing
 disable_first_popup = false
 
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -562,6 +562,7 @@ void Config::Load()
   spdlog::debug("");
 
   this->disable_escape_exit    = get_config_or_default(config, parsed, "ui", "disable_escape_exit", DCU::disable_escape_exit, write_config);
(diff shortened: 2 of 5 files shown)


Guffawaffle:
## QA Checklist

### Prerequisites
- [ ] Built with `xmake build stfc-community-patch` (releasedbg mode)
- [ ] Fresh game launch (no leftover state)

### Escape Behavior — `disable_escape_exit = false` (default)
- [ ] Escape key works normally (opens exit prompt)
- [ ] `escape_exit_timer` setting has no effect

### Escape Behavior — `disable_escape_exit = true`, `escape_exit_timer = 0`
- [ ] Single Escape press is fully swallowed (no exit prompt)
- [ ] Rapid double-tap Escape is also swallowed (timer disabled)
- [ ] This matches existing v1.0.0 behavior exactly

### Escape Behavior — `disable_escape_exit = true`, `escape_exit_timer = 500`
- [ ] Single Escape press is swallowed
- [ ] Double-tap Escape within ~500ms triggers the native exit prompt
- [ ] Double-tap Escape with >500ms gap does NOT trigger exit (second press starts new window)
- [ ] Works correctly from station view, system view, and galaxy view

### Config Loading
- [ ] `escape_exit_timer` appears in `community_patch_runtime.vars` under `[ui]`
- [ ] Missing `escape_exit_timer` in TOML defaults to `0` (backward compatible)
- [ ] `example_community_patch_settings.toml` documents the new option

### Regression
- [ ] Escape still closes open panels/viewers when `disable_escape_exit = true`
- [ ] F10 hard kill still works regardless of settings
- [ ] No crashes during rapid Escape key mashing

netniV:
Code looks good, can see a reason for this, not sure I'd ever use it because I spam escape so it would get triggered, and that dastardly dialog would trigger me.  At least by default it maintains the standard functionality, should probably add a comment that statess the same ms value for trigger as the double tap locate.

Guffawaffle:
Updated the comments to be properly user-facing — no more references to internal `select_timer`. The TOML now reads:

```toml
# When disable_escape_exit is true, this lets you double-tap Escape to still
# get the exit prompt. The value is the maximum gap in milliseconds between
# two Escape presses to count as a double-tap.
#
#   0   = double-tap disabled, Escape is fully blocked (default)
#   500 = press Escape twice within half a second to trigger the exit prompt
#
# Has no effect when disable_escape_exit is false.
escape_exit_timer = 0
```

Force-pushed in 278f419.

PR #126, opened 2026-04-13
Feature: hotkey dispatch table + 3-tier architecture extraction

Guffawaffle:
## Summary

Two-phase refactor of the hotkey system:

1. **Dispatch table** - Replaces the ~90-line if/else cascade with a table-driven dispatch loop
2. **3-tier architecture extraction** - Breaks the 800-line `hotkeys.cc` monolith (33 includes) into a layered architecture, reducing TU bloat in the thin SPUD hook file

Both phases are intended as behavior-preserving refactors. This PR now contains only the hotkey refactor scope; no sync or build-system changes are included.

---

## Phase 1: Dispatch Table

The monolithic if/else chain that checked ~43 GameFunction bindings one-by-one has been extracted into:

- **`hotkey_dispatch.h`** - Core types:
  - `DispatchDecision` enum: `NoMatch`, `HandledStop`, `HandledAllowOriginal`
  - `InputMode` enum: `Down` (single trigger) vs `Pressed` (repeat-while-held)
  - `HotkeyEntry` struct: `GameFunction` + function pointer handler + `InputMode`

- **`hotkey_dispatch.cc`** - 43 individual handler functions + static immutable dispatch table
  - Section navigation (22 GotoSection + 2 ChangeNavigationSection + ShowBookmarks)
  - UI scale adjustment (4, using `InputMode::Pressed`)
  - Config toggles (7)
  - Log level changes (6)
  - Ship management (ShowShips)

The if/else chain is replaced with a 12-line dispatch loop:
```cpp
for (const auto& entry : GetHotkeyDispatchTable()) {
  bool active = (entry.input_mode == InputMode::Pressed)
                  ? MapKey::IsPressed(entry.game_function)
                  : MapKey::IsDown(entry.game_function);
  if (active) {
    auto decision = entry.handler();
    if (decision == DispatchDecision::HandledStop) return;
    break;
  }
}
```

This is the structural foundation for future key fallthrough support - the dispatch table makes it straightforward to add per-entry properties without touching the main hook logic.

The table is now static immutable storage rather than a runtime-built vector, which removes the startup init/allocation step from hook installation.

---

## Phase 2: 3-Tier Architecture Extraction

Extracts all remaining procedural logic from `hotkeys.cc` into a layered architecture:

### Layer 1 - Thin Hooks (`hotkeys.cc`, ~120 lines, 5 includes)
SPUD detour hooks only, delegating all logic to the router. This keeps the detour TU materially smaller than the previous monolith.

### Layer 2 - Dispatcher (`hotkey_router.cc/h`)
Full orchestration logic from `ScreenManager_Update_Hook`:
- Key cache reset + enable/disable toggle
- Ship selection routing (double-tap timing, shift-tow)
- Dispatch table loop
- Chat open/switch logic
- Escape + viewer hiding + exit suppression
- Space action execution
- Info pending multi-frame countdown

### Layer 3 - Concern Files
Each file owns a focused domain with its own state:

| File | Responsibility | Key Functions |
|---|---|---|
| `fleet_actions.cc/h` | Ship selection, space actions, fleet recall/repair | `HandleShipSelection`, `ExecuteSpaceAction`, `DidExecuteRecall`, `DidExecuteRepair`, `GetHullTypeFromBattleTarget` |
| `viewer_mgmt.cc/h` | Object viewer visibility, rewards dismiss, info pending | `CanHideViewers`, `DidHideViewers`, `TryDismissRewardsScreen`, `HandleActionView`, `TickInfoPending` |
| `cargo_display.cc/h` | Auto-show cargo on target scan | `CheckShowCargo`, `HandleCargoBindContext`, `HandleCargoShowFleet` |
| `navigation.cc/h` | Section navigation, officer canvas | `GotoSection`, `ChangeNavigationSection`, `MoveOfficerCanvas` |

### Header fixes (implicit dependency cleanup)
- `PreScanTargetWidget.h`: added missing `ObjectViewerBaseWidget.h` and `ScanEngageButtonsWidget.h` includes
- `hotkey_dispatch.cc`: replaced forward declarations with `#include "patches/navigation.h"`

---

## File Impact Summary

| File | Change | Lines |
|---|---|---|
| `hotkeys.cc` | Rewritten: 800 -> ~120 lines, 33 -> 5 includes | -703 |
| `hotkey_router.cc/h` | **New** - Layer 2 dispatcher | +257 |
| `fleet_actions.cc/h` | **New** - Ship selection + space actions | +352 |
| `viewer_mgmt.cc/h` | **New** - Viewer management + info pending | +157 |
| `cargo_display.cc/h` | **New** - Cargo auto-show | +61 |
| `navigation.cc/h` | **New** - Section navigation | +43 |
| `hotkey_dispatch.cc` | Updated: uses navigation.h | +3/-2 |
| `PreScanTargetWidget.h` | Fixed: added missing includes | +2 |

## Validation

- Rebased onto `origin/main` to remove the unrelated sync/LTO commit from this PR
- Verified the PR file list no longer includes `mods/src/patches/parts/sync.cc`, `mods/xmake.lua`, `.gitignore`, or `.vscode/settings.json`
- Converted the dispatch table from a runtime-built vector to static immutable storage and removed the router init step
- Rebuilt the branch with `xmake build mods` after the static-table change


netniV:
Whilst I like the idea of this, I'm just conscious of two things, functional calling overhead, and ensuring that the originally defined priorities are still correct because there are some funny loops that occur to make certain things happen in different frames... as well as dropping through from one to another so testing is definitely required to make sure the edge cases still work.

Guffawaffle:
ensuring that the originally defined priorities are still correct <-- this feels important and something I wouldn't be able to easily grok. I can definitely come up with some tests, though, to see about call overhead and such. I know I ran into some problems with this and we definitely don't want to slow down the game. Being on a pretty powerful machine makes that hard to test for me... I bet I can run the process constrained to certain performance characteristic though if I dig into it.

Guffawaffle:
also I know there were like, two classes when I had dove into this originally as I would define them. One is simple redirecting the keypress but the others are like -- if this framee is here do that but if it was something else first do the other (double taps, chat iirc, etc). In theory this covers both cases but I'll make also try to identify and test the more complex interactions.

Guffawaffle:
Addressed the dispatch-table overhead concern in a narrow follow-up commit.

Changes:
- switched the table from a runtime-built `std::vector` to static immutable storage
- removed the separate router init step that only existed to build that table
- kept the priority order and dispatch loop shape unchanged

Why:
- the table is fixed, small, and read-only, so runtime construction was extra machinery with no upside here
- this removes a startup allocation/copy path and one init foot gun without changing the hotkey ordering model

What this does **not** claim to solve:
- it trims setup overhead and simplifies the design, but it does not by itself prove the frame-order edge cases are correct; those still need behavioral testing

Validation:
- `xmake build mods`

PR #137, opened 2026-04-20
feat: OS-level notifications for game events (battles, fleet arrivals)

Guffawaffle:
## Summary

Adds a platform-abstracted notification service that delivers OS-level (Windows toast) notifications for configurable in-game event types. Defaults to **off** — no user-visible behavior change unless `notify_banner_types` is set.

---

### Notification Service (`notification_service.h/.cc`)

- WinRT-based toast delivery (zero third-party deps — no WinToastLib)
- Hard `#if _WIN32` guard: non-Windows platforms are compile-time no-ops
- `notification_init()` — idempotent init, safe to call from any hook installer
- `notification_show(title, body)` — fire-and-forget WinRT toast
- `notification_handle_toast(toast)` — called from existing `EnqueueToast`/`EnqueueOrCombineToast` hooks; resolves the toast's `LTC` text via `LanguageManager::Localize` before deciding whether to notify

### Config

```toml
[ui]
# Comma-separated list using the same type names as disabled_banner_types.
# Empty string = no notifications (default).
notify_banner_types = ""
```

### Battle Result Parser (`battle_notify_parser.cc/h`)

Extracts player/enemy names and ship hull classes from `BattleResultHeader` for battle toast types (`Victory`, `Defeat`, `PartialVictory`, `Station*`, `FleetBattle`, `Armada*`, `Assault*`). All IL2CPP property accesses wrapped in SEH guards for crash safety.

Format: `Guffawaffle (Newton) vs Assimilated Coryn-class`

- NPC profiles have empty display names — hull class name used as identifier instead of `NPC#<locaId>`
- Non-battle toasts fall back to the LTC text resolver

New prime headers: `BattleResultHeader.h`, `UserProfile.h`, `SpecService.h`

### Fleet Arrival Detection (`parts/fleet_arrival.cc`)

Hooks `FleetStateWidget::SetWidgetData` and watches `FleetPlayerData` state transitions. Fires a notification on `Warping → Impulsing` (ship dropped out of warp into destination system).

Uses `SPUD_STATIC_DETOUR` — no MinHook dependency.

---

## Notes

- All notification calls pass through the existing toast hook path (`disable_banners.cc`) — notification delivery is checked *before* the banner-disable filter, so players can receive an OS notification even for banners they've suppressed in-game
- The `notify_banner_types` config uses the same type names as `disabled_banner_types`, so power users can mirror their suppressed banners as OS toasts
- `fleetarrivalhooks` toggle added to `defaultconfig.h` for debug-build gating

### DIFF
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -324,6 +324,12 @@ disable_toast_banners = false
 
 disabled_banner_types = ""
 
+# System notifications for in-game events. Comma-separated list of toast types
+# to deliver as OS-level notifications (Windows toast). Uses the same type names
+# as disabled_banner_types. Empty string = no notifications.
+# Example: "Victory, Defeat, IncomingAttack, StationBattle"
+notify_banner_types = ""
+
 # Subgroup: Chat
 # --------------
 
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -689,6 +689,29 @@ void Config::Load()
 
   parsed["ui"].as_table()->insert_or_assign("disabled_banner_types", bannerString);
 
+  // Parse notify_banner_types using the same bannerTypes lookup table
+  auto notify_banner_types_str =
+      get_config_or_default<std::string>(config, parsed, "ui", "notify_banner_types", DCU::notify_banner_types, write_log);
+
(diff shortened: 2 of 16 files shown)


Guffawaffle:
Closed for fixing due to rebase back to SPUD has caused unintended drift.

PR #145, opened 2026-04-27
feat: add prerelease and self-update flow

a5ehren:
given the number of users that are either not in the discord or unable to dig a fix out of the actions builds, I thought it might be a good idea to start tagging pre-releases from `dev` on every merged PR. 

Then the obvious next step is to add auto-updaters to the MacOS launcher and Windows version.dll. They default to stable, versioned releases only, but can be configured to use the tagged prereleases instead.


netniV (approved):

netniV:
Not sure I want this on every PR.  I can certainly see a thing for a daily or weekly flow.  Maybe some more discussion about why it's needed?

Would love a self-update, though I do wonder if that's possible with Windows thinking that the dll is in use (unlike Linux).

a5ehren:
I can edit to whatever we like, sure. 

The windows side should work by writing out a little powershell script that kills the process to unload the DLL

netniV:
> I can edit to whatever we like, sure.
> 
> The windows side should work by writing out a little powershell script that kills the process to unload the DLL



> I can edit to whatever we like, sure.
> 
> The windows side should work by writing out a little powershell script that kills the process to unload the DLL

If it was to do that, can only be at startup because any other time, we have no idea what they are doing which will annoy players and we're in the business of trying to help them.... :D 

We'd also need to make sure we aren't in the process of some delta sync that gets lost.

Guffawaffle:
I'm wondering if we would want to look into an Installer and Launcher that could offer a settings UI as well as an Update button. It could be thin and part of an optional download. This would give users who are accustomed to the current experience 0 friction or change in behavior while allowing those who would be interested to download this.

Similar experiences exist for World of Warcraft addons such as TukUi (ElvUi) and the paid Zygor Guides app.

a5ehren:
> > I can edit to whatever we like, sure.
> > The windows side should work by writing out a little powershell script that kills the process to unload the DLL
> 
> > I can edit to whatever we like, sure.
> > The windows side should work by writing out a little powershell script that kills the process to unload the DLL
> 
> If it was to do that, can only be at startup because any other time, we have no idea what they are doing which will annoy players and we're in the business of trying to help them.... :D
> 
> We'd also need to make sure we aren't in the process of some delta sync that gets lost.

Yeah, the check only happens at startup before the game DLL gets loaded. After the check everything is the same.

a5ehren:
> I'm wondering if we would want to look into an Installer and Launcher that could offer a settings UI as well as an Update button. It could be thin and part of an optional download. This would give users who are accustomed to the current experience 0 friction or change in behavior while allowing those who would be interested to download this.
> 
> Similar experiences exist for World of Warcraft addons such as TukUi (ElvUi) and the paid Zygor Guides app.

I'll vibe-code something that looks like the mac launcher, probably in another PR?

netniV:
> > I'm wondering if we would want to look into an Installer and Launcher that could offer a settings UI as well as an Update button. It could be thin and part of an optional download. This would give users who are accustomed to the current experience 0 friction or change in behavior while allowing those who would be interested to download this.
> > Similar experiences exist for World of Warcraft addons such as TukUi (ElvUi) and the paid Zygor Guides app.
> 
> I'll vibe-code something that looks like the mac launcher, probably in another PR?

Let me add you to a project that I started to work on but haven't touched in two or three months lol.

netniV:
https://github.com/netniV/stfc-mod/actions/runs/25500288946/job/74831083297 -- seems to have issues with using a default configuration namespace that has since been removed ...

a5ehren:
> https://github.com/netniV/stfc-mod/actions/runs/25500288946/job/74831083297 -- seems to have issues with using a default configuration namespace that has since been removed ...

oh whoops, lazy merge reading

PR #147, opened 2026-04-28
update prime headers with newest dump

a5ehren:
I'm not sure how you normally do this, but I'm testing a new workflow


netniV (commented):
At a quick glance, there are a number of functions removed which appear to be ones that are utilised in other areas.  Are we sure these should be removed?  Where they a copy and paste hangover originally?

a5ehren:
Yeah I have not done any testing yet besides compiling the dylib on mac. I'll keep this as a draft until I do that.

PR #149, opened 2026-04-29
Align local macOS packaging with CI

wycats:
This makes the local macOS build script assemble the same app bundle layout that CI packages.

Why:
- I was testing the macOS launcher locally, but the dev script only launched the raw `macOSLauncher.app` output.
- CI adds the auxiliary loader, injected dylib, generated `Info.plist`, icon, and signs the final `STFC Community Mod.app` bundle.
- Mirroring that locally makes macOS testing closer to the release artifact.

Changes:
- Have `mac-build-test-debug.sh` package and verify `STFC Community Mod.app` after building.
- Keep xmake's `macOSLauncher.app` in place for incremental rebuilds, and copy it to the packaged app instead of moving it.
- Launch the packaged app for launcher runs.
- Match `create-mac-dmg.sh`'s deployment target to CI / plist (`13.5`).

I have a few separate macOS fixes from local testing, but I'm sending them as small focused PRs rather than one large batch. Happy to adjust the script shape if you'd prefer this logic to live elsewhere.

### DIFF
--- a/scripts/mac-build-test-debug.sh
+++ b/scripts/mac-build-test-debug.sh
@@ -216,20 +216,27 @@ package_app_bundle() {
         exit 1
     fi
 
+    local icon_path="${PROJECT_ROOT}/assets/launcher.icns"
+    if [[ ! -f "$icon_path" ]]; then
+        print_error "Launcher icon not found at: $icon_path"
+        exit 1
+    fi
+
+    local info_plist_path="${PROJECT_ROOT}/macos-launcher/src/Info.plist"
+    if [[ ! -f "$info_plist_path" ]]; then
+        print_error "Generated Info.plist not found at: $info_plist_path"
+        print_info "Run the configure/build step so xmake can generate it from Info.plist.template"
+        exit 1
+    fi
+
     rm -rf "$STFC_APP_PATH"
     ditto "$APP_PATH" "$STFC_APP_PATH"
 
     mkdir -p "${STFC_APP_PATH}/Contents/Resources"
     cp "$LOADER_PATH" "${STFC_APP_PATH}/Contents/stfc-community-mod-loader"
     cp "$dylib_path" "${STFC_APP_PATH}/Contents/libstfc-community-mod.dylib"
-
-    if [[ -f "${PROJECT_ROOT}/assets/launcher.icns" ]]; then
-        cp "${PROJECT_ROOT}/assets/launcher.icns" "${STFC_APP_PATH}/Contents/Resources/"
-    fi
-
-    if [[ -f "${PROJECT_ROOT}/macos-launcher/src/Info.plist" ]]; then
-        cp "${PROJECT_ROOT}/macos-launcher/src/Info.plist" "${STFC_APP_PATH}/Contents/"
-    fi
+    cp "$icon_path" "${STFC_APP_PATH}/Contents/Resources/"
+    cp "$info_plist_path" "${STFC_APP_PATH}/Contents/"
 
     print_info "Code signing packaged application..."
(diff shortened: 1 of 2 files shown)


scripts/mac-build-test-debug.sh
```diff
@@ -188,6 +188,54 @@ build_project() {
...
+    cp "$LOADER_PATH" "${STFC_APP_PATH}/Contents/stfc-community-mod-loader"
+    cp "$dylib_path" "${STFC_APP_PATH}/Contents/libstfc-community-mod.dylib"
+
+    if [[ -f "${PROJECT_ROOT}/assets/launcher.icns" ]]; then
+        cp "${PROJECT_ROOT}/assets/launcher.icns" "${STFC_APP_PATH}/Contents/Resources/"
+    fi
+
+    if [[ -f "${PROJECT_ROOT}/macos-launcher/src/Info.plist" ]]; then
+        cp "${PROJECT_ROOT}/macos-launcher/src/Info.plist" "${STFC_APP_PATH}/Contents/"
+    fi
```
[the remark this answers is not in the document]
wycats: Addressed in c538fce: local packaging now fails if the icon or generated Info.plist is missing, matching CI instead of silently producing an incomplete bundle.

scripts/mac-build-test-debug.sh
```diff
@@ -242,7 +285,7 @@ run_app() {
     if [[ "$USE_LAUNCHER" == true ]]; then
         print_info "Launching application with launcher..."
         print_info "Crash dumps will be generated at: ~/Library/Logs/DiagnosticReports/"
-        open "$APP_PATH"
+        open "$STFC_APP_PATH"
         
         print_success "Launcher launched"
         print_info "To view logs, use: log stream --predicate 'process == \"macOSLauncher\"' --level debug"
```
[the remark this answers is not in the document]
wycats: Addressed in c538fce: launcher debug now targets the packaged STFC Community Mod.app executable, so run/debug use the same bundle layout.

a5ehren (approved):

scripts/mac-build-test-debug.sh
```diff
@@ -331,7 +381,7 @@ debug_app() {
     local exec_name
     
     if [[ "$USE_LAUNCHER" == true ]]; then
-        executable="${APP_PATH}/Contents/MacOS/macOSLauncher"
+        executable="${STFC_APP_PATH}/Contents/MacOS/macOSLauncher"
```
netniV: Pretty sure this is wrong... it should be the APP path not the game path....

wycats: I get why you think it's sus. In this script, `APP_PATH` is the raw `xmake` output (`macOSLauncher.app`), and `STFC_APP_PATH` is the packaged `STFC Community Mod.app` bundle this PR creates.

I changed the debug path to match the launcher run path, which opens `STFC_APP_PATH`.

That said, the variable name reads like the actual game app path, so I can see why it's confusing. What do you think about renaming it to something like `PACKAGED_APP_PATH` / `COMMUNITY_MOD_APP_PATH` to make it clearer?

wycats: I pushed the above as a tweak. Lemme know what you think.

wycats:
I pushed a small rename: `STFC_APP_PATH` is now `PACKAGED_APP_PATH`. 

The behavior is unchanged. Launcher run/debug both use the packaged `STFC Community Mod.app`, but the old variable name made it look like it might refer to the actual game app path, which I agree was confusing.

I think this should be safe to merge now?

PR #150, opened 2026-04-29
Handle unsigned macOS game executable signing

wycats:
This handles a macOS launcher case where the game executable is unsigned.

Why:
- PR #123 avoids re-signing when the game already has the entitlements needed for DYLD-based mod loading.
- While testing locally, I hit a different state: the game executable reported `code object is not signed at all`.
- The existing fallback then tried to sign the executable path directly, but `codesign` walked the app bundle and failed on an invalid nested `NativeScript.bundle`.

Changes:
- Distinguish valid, missing, and unsigned entitlement states.
- For unsigned executables, sign a temporary copy and write the signed bytes back to the game executable.
- Preserve the executable's original permissions.
- Keep the existing user-confirmed file-picker flow for the signed-but-missing-entitlements case.

Validation:
- Restored an unsigned copy of the game executable.
- Clicked Engage from the launcher.
- Launcher log showed `Game executable is unsigned`, then `Successfully applied loader entitlements to unsigned game executable`.
- Game launched and the mod loaded.
- `./scripts/mac-build-test-debug.sh -m release build`

I split this from the local build-script PR so the signing behavior can be reviewed independently.

### DIFF
--- a/macos-launcher/src/PatchEntitlements.swift
+++ b/macos-launcher/src/PatchEntitlements.swift
@@ -33,6 +33,12 @@ enum EntitlementError: Error {
     }
 }
 
+enum EntitlementCheckResult {
+    case valid
+    case missing
+    case unsigned
+}
+
 /// Ensures the game has the required loader entitlements, applying them only if missing.
 /// The game's original Scopely signature already includes the entitlements we need,
 /// so this typically just verifies and returns without modifying anything.
@@ -44,15 +50,45 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
         throw EntitlementError.gamePathNotFound
     }
 
+    let gameAppPath = (gamePath as NSString).deletingLastPathComponent
+        .replacingOccurrences(of: "/Contents/MacOS", with: "")
+
     // Always verify entitlements first — even after a game update, the new binary
     // typically ships with the entitlements we need from Scopely's signature.
-    if checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) {
+    switch checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) {
+    case .valid:
         logger.info("Game already has the required loader entitlements, no re-signing needed")
         // Clear any pending re-application flag since entitlements are valid
         if UserDefaults.standard.bool(forKey: "forceEntitlementReapplication") {
             UserDefaults.standard.set(false, forKey: "forceEntitlementReapplication")
         }
         return
+    case .unsigned:
+        logger.info("Game executable is unsigned, re-signing is needed")
+        if isGameRunning(bundlePath: gameAppPath) {
+            logger.error("Game is currently running, cannot modify signature")
+            throw EntitlementError.entitlementApplicationFailed
+        }
+
+        let applySuccess = applyEntitlementsDirectly(
+            appBundlePath: gameAppPath,
+            mainExecutablePath: gamePath,
+            entitlements: loaderEntitlements,
+            signingIdentity: signingIdentity
+        )
+        guard applySuccess else {
+            throw EntitlementError.entitlementApplicationFailed
+        }
+
+        guard case .valid = checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
+            throw EntitlementError.entitlementVerificationFailed
+        }
+
+        logger.info("Successfully applied loader entitlements to unsigned game executable")
+        UserDefaults.standard.set(false, forKey: "forceEntitlementReapplication")
+        return
+    case .missing:
+        break
     }
 
     // Entitlements are missing — fall back to re-signing.
@@ -60,8 +96,6 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
     logger.warning("Game is missing required loader entitlements, re-signing is needed")
 
     // Check if the game is currently running
-    let gameAppPath = (gamePath as NSString).deletingLastPathComponent
-        .replacingOccurrences(of: "/Contents/MacOS", with: "")
     if isGameRunning(bundlePath: gameAppPath) {
         logger.error("Game is currently running, cannot modify signature")
         throw EntitlementError.entitlementApplicationFailed
@@ -103,7 +137,7 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
     }
 
     // Verify the entitlements were applied successfully to the main executable
-    guard checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
+    guard case .valid = checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
         throw EntitlementError.entitlementVerificationFailed
     }
 
@@ -158,8 +192,8 @@ func fetchGamePath() throws -> String? {
 /// - Parameters:
 ///   - appPath: Path to the .app bundle or executable to check
 ///   - expectedEntitlements: Dictionary of entitlement keys and their expected values
-/// - Returns: true if all expected entitlements match, false otherwise
-func checkEntitlements(appPath: String, expectedEntitlements: [String: Any]) -> Bool {
+/// - Returns: valid if all expected entitlements match, unsigned if the app has no signature, missing otherwise
+func checkEntitlements(appPath: String, expectedEntitlements: [String: Any]) -> EntitlementCheckResult {
     // Create process to run codesign command
     let process = Process()
     process.executableURL = URL(fileURLWithPath: "/usr/bin/codesign")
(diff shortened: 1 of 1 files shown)


macos-launcher/src/PatchEntitlements.swift
```diff
@@ -176,49 +210,53 @@ func checkEntitlements(appPath: String, expectedEntitlements: [String: Any]) ->
...
+                if errorString.contains("code object is not signed at all") {
+                    logger.info("Game executable is unsigned")
+                    return .unsigned
+                }
+                logger.error("codesign failed with status \(process.terminationStatus)")
                 logger.error("codesign error output: \(errorString)")
             }
-            return false
+            return .missing
         }
```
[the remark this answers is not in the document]
wycats: Addressed in e128672: checkEntitlements now has a distinct error result for codesign inspection failures, and the launcher surfaces that as verification failure instead of entering the re-signing path.

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -447,6 +466,18 @@ func executeCodesignDirectly(
             return false
         }
 
+        do {
+            let attributes = try fileManager.attributesOfItem(atPath: mainExecutablePath)
+            let signedData = try Data(contentsOf: tempExecutable)
+            try signedData.write(to: URL(fileURLWithPath: mainExecutablePath), options: .atomic)
+            if let permissions = attributes[.posixPermissions] {
+                try fileManager.setAttributes([.posixPermissions: permissions], ofItemAtPath: mainExecutablePath)
+            }
```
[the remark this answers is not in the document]
wycats: Addressed in e128672: the signed temporary executable is now swapped in with FileManager.replaceItemAt instead of reading the whole binary into memory and writing it back manually.

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -176,49 +210,53 @@ func checkEntitlements(appPath: String, expectedEntitlements: [String: Any]) ->
...
-            return false
+            return .missing
         }
 
         // Parse the plist
         guard let plist = try? PropertyListSerialization.propertyList(from: data, format: nil),
               let entitlementsDict = plist as? [String: Any] else {
             logger.error("Failed to parse entitlements plist")
-            return false
+                        return .missing
```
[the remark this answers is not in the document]
wycats: Addressed in e128672: fixed the indentation in that guard block.

a5ehren:
How did you hit a state where the game app is unsigned?

wycats:
> How did you hit a state where the game app is unsigned?

I honestly don't know but it's what happened. I can try to provide more logs or other information. I agree it's weeeeird.

a5ehren:
Nah it's fine. It's a decent edge case to have coverage for

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -44,24 +51,54 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
...
+
+        guard case .valid = checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
+            throw EntitlementError.entitlementVerificationFailed
+        }
+
+        logger.info("Successfully applied loader entitlements to unsigned game executable")
+        UserDefaults.standard.set(false, forKey: "forceEntitlementReapplication")
+        return
+    case .missing:
+        break
```
netniV, line 92: If an entitlement key is missing, nothing happens?  Yet the code returning this breaks out of the loop on the first missing entitlement.... surely we should be saying something somewhere?

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -368,51 +409,32 @@ func executeCodesignDirectly(
...
-       !errorData.isEmpty,
-       let errorString = String(data: errorData, encoding: .utf8), !errorString.isEmpty {
-        logger.debug("Remove signature stderr: \(errorString)")
+    defer {
+        try? fileManager.removeItem(at: tempExecutable)
     }
 
-    // Now sign just the main executable with the entitlements
-    // This preserves nested code signatures and only modifies what we need
-    logger.info("Signing executable with entitlements...")
```
netniV, line 402: Do we care that a bunch of debug info was removed?

netniV (changes requested):

macos-launcher/src/PatchEntitlements.swift
a5ehren: Reviewer note: the concern is valid, though partly a readability issue. Two things:

- `break` here just exits the `switch` (Swift `break`, not a loop break), so control *does* fall through to the re-signing block below, which logs `logger.warning("Game is missing required loader entitlements, re-signing is needed")`. So it isn't a no-op — but a bare `case .missing: break` with no comment reads like one, which is what makes this confusing.
- The genuinely weak part: *which* entitlement is missing is only logged at `.debug` (in `checkEntitlements`, the "not found"/"mismatch" branches). On a default logger that's invisible, so when the re-sign path runs you can't tell *why* from normal logs.

Suggested fix — make the fall-through explicit and surface the missing key at a visible level:

```swift
case .missing:
    // Fall through to the re-signing flow below.
    break
```
and bump the `checkEntitlements` diagnostics from `.debug` to `.notice` so the missing/mismatched key shows up in logs.

macos-launcher/src/PatchEntitlements.swift
a5ehren: Reviewer note: mostly fine, one line worth restoring.

- The `--remove-signature` stdout/stderr logging is gone because the *step itself* is gone — the new copy-then-`--force`-sign approach never removes a signature first, so that logging is correctly obsolete. No loss.
- The explicit `logger.info("Command: codesign --force --sign ... \(mainExecutablePath)")` line, however, was genuinely useful for diagnosing signing failures and is gone. The sign process's stdout/stderr is still logged, so you're not blind, but you lose the exact invocation.

Given this PR exists *because* of a hard-to-reproduce signing edge case, keeping that diagnostic has real value. Suggest restoring an equivalent line before `signProcess.run()`:

```swift
logger.info("Command: codesign --force --options runtime --sign \(signingIdentity) --entitlements \(entitlementsPath) \(tempExecutable.path)")
```

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -44,24 +51,54 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
...
-    if checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) {
+    switch checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) {
+    case .valid:
         logger.info("Game already has the required loader entitlements, no re-signing needed")
         // Clear any pending re-application flag since entitlements are valid
         if UserDefaults.standard.bool(forKey: "forceEntitlementReapplication") {
             UserDefaults.standard.set(false, forKey: "forceEntitlementReapplication")
         }
         return
+    case .unsigned:
```
a5ehren, line 67: Behavioral divergence worth confirming is intentional: the `.unsigned` path calls `applyEntitlementsDirectly` directly, skipping the `requestGameBundleAccess` file picker that the `.missing` path uses for explicit user consent. Since the launcher is **not** sandboxed (`app-sandbox = false` in macOSLauncher.entitlements), file access still works via the App Management TCC grant, so this isn't a correctness bug — but one path asks "may I modify your game binary?" and the other silently rewrites it. Note the doc comments on `applyEntitlementsDirectly`/`executeCodesignDirectly` still say "Must be called while security-scoped resource access is active", which this caller now violates. Either route unsigned through the same consent flow or relax those comments.

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -44,24 +51,54 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
...
+            throw EntitlementError.entitlementVerificationFailed
+        }
+
+        logger.info("Successfully applied loader entitlements to unsigned game executable")
+        UserDefaults.standard.set(false, forKey: "forceEntitlementReapplication")
+        return
+    case .missing:
+        break
+    case .error:
+        throw EntitlementError.entitlementVerificationFailed
```
a5ehren, line 94: Behavior change to confirm: previously any falsy `checkEntitlements` result (codesign failure, empty output, parse failure) fell through to the re-signing path. Now `.error` throws `entitlementVerificationFailed` immediately and never attempts the file-picker fallback. This is arguably more correct (don't modify the binary when you couldn't even read entitlements), but it's a real change for transient codesign failures the old code would have recovered from. Intentional?

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -103,7 +140,7 @@ func ensureGameHasLoaderEntitlements(signingIdentity: String = "-") throws {
     }
 
     // Verify the entitlements were applied successfully to the main executable
-    guard checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
+        guard case .valid = checkEntitlements(appPath: gamePath, expectedEntitlements: loaderEntitlements) else {
```
a5ehren, line 143: Indentation glitch: this `guard` is over-indented relative to the function body. No SwiftLint/swift-format config exists in `macos-launcher/`, so CI won't catch it — worth fixing by hand for consistency.

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -176,49 +213,53 @@ func checkEntitlements(appPath: String, expectedEntitlements: [String: Any]) ->
...
-            return false
+            return .error
         }
 
         // Parse the plist
         guard let plist = try? PropertyListSerialization.propertyList(from: data, format: nil),
               let entitlementsDict = plist as? [String: Any] else {
             logger.error("Failed to parse entitlements plist")
-            return false
+                        return .error
```
a5ehren, line 239: Indentation glitch: `return .error` is over-indented compared to the surrounding block. Align with the other statements in this `guard` body.

a5ehren (commented):
A few additional findings from a fresh review pass, alongside the replies on netniV's threads. None are blocking on their own, but I'd want answers on the `.error` fall-through change and the unsigned-path consent divergence before merge.

macos-launcher/src/PatchEntitlements.swift
```diff
@@ -447,6 +469,14 @@ func executeCodesignDirectly(
             return false
         }
 
+        do {
+            let executableURL = URL(fileURLWithPath: mainExecutablePath)
+            try fileManager.replaceItemAt(executableURL, withItemAt: tempExecutable)
```
a5ehren, line 474: `replaceItemAt` (without `.usingNewMetadataOnly`) is what actually backs the PR's "preserve original permissions" claim — it preserves the destination's metadata. That's load-bearing and easy to silently break later by adding the wrong option. Worth a one-line comment here noting the intent so a future change doesn't regress it.

netniV:
If you can address these @wycats we can look at getting this merged in.

PR #154, opened 2026-04-30
Add raw static EntityGroup capture

wycats:
## Summary

Adds an opt-in local developer capture path for raw static/spec `EntityGroup` payloads.

When enabled, the mod writes raw protobuf bytes to a local content-addressed blob store and appends JSONL manifest entries with:

- timestamp
- `EntityGroup` numeric type
- type name
- byte count
- SHA-256
- relative blob path
- whether the blob was newly written or already present

This is intended as infrastructure for later static catalog extraction work. It does not parse the payloads and does not send them to sync targets.

## Configuration

Adds `[sync.raw_static_capture]`:

- `enabled = false`
- `directory = ""`
- `max_bytes = 52428800`
- `include_types = ""`

`include_types` is an optional comma-separated numeric `EntityGroup` allowlist. Empty captures all known static/spec/config types.

## Scope

- Local-only raw capture
- Static/spec/config allowlist only
- No HTTP upload
- No dynamic player/account data capture
- No typed decoding yet
- No localization/assets extraction yet

## Validation

- `git diff --check`
- `./scripts/mac-build-test-debug.sh -m release build`
- macOS runtime smoke with capture enabled wrote `manifest.jsonl` plus content-addressed blobs under `/Users/wycats/Desktop/stfc-static-capture-smoke`
  - 183 manifest entries
  - 92 unique blob files
  - sample captured types included `BlueprintSpecs`, `StarbaseSpecs`, `ResourceSpecs`, `HazardSpecs`, and `OfficerLevelRewardsSpecs`
  - verified a captured blob path hash with `shasum -a 256`

### DIFF
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -450,3 +450,17 @@ verify_ssl = true
 
 # The url of the server to send data to
 # url = ""
+
+#  <[========================================]>
+#       Raw static capture
+#  <[========================================]>
+
+# Optional developer tool for capturing raw static/spec EntityGroup protobuf payloads locally.
+# Disabled by default. This does not send data to sync targets and does not parse the payloads.
+# The directory receives content-addressed raw blobs plus a manifest.jsonl index.
+# include_types is a comma-separated numeric EntityGroup type allowlist. Empty means all known static/spec types.
+[sync.raw_static_capture]
+enabled = false
+directory = ""
+max_bytes = 52428800
+include_types = ""
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -23,6 +23,7 @@ namespace DCC = DefaultConfig::Control;
 namespace DCU = DefaultConfig::UI;
 namespace DCBS = DefaultConfig::Buffs;
 namespace DCS = DefaultConfig::Sync;
+namespace DCSR = DefaultConfig::Sync::RawStaticCapture;
 namespace DCSC = DefaultConfig::SystemConfig;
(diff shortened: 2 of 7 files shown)


a5ehren:
What is the use case here? We don't really have any reason to expose this in a public mod, seems more like a development tool for something else.

wycats:
Addressed the Copilot comments in `1e6947c`:

- changed SHA-256 hashing to process full 64-byte blocks directly and only keep a small final padding buffer instead of copying the full payload into a padded vector
- checked and reported `std::filesystem::exists` errors before deciding whether a blob is missing
- write new blobs through a temporary file and rename into place, cleaning up the temp file on failures

Validation:
- `git diff --check`
- `./scripts/mac-build-test-debug.sh -m release build`
- SHA-256 vectors for empty string, `abc`, and one million `a` characters

wycats:
> What is the use case here? We don't really have any reason to expose this in a public mod, seems more like a development tool for something else.

Yeah, I made this a draft because I wasn't sure whether it belongs here. I could see a world where this mod consolidated all kinds of data dumping in one place (so the community only had to maintain the core infra for a single mod). I could also see a world where this is clearly out of scope.

I need the data for another stfc-related project I'm working on, and it was easy to build in terms of this mod, but I didn't want to assume anything.

wycats:
I'm thinking I'll close this and think about whether it belongs here at all, and if it does, if it can be accomplished with a lighter-touch way to plug in external "plugins" or something like that.

PR #160, opened 2026-05-02
Add opt-in key parsing regression tests

Guffawaffle:
## Reviewer Quick Path

This is a draft/prototype PR. No merge review is requested until it is reworked.

Known public state:

- `netniV` requested that the tests use the main implementation rather than moving the constant into the test file
- the latest GitHub `build-win` check is failing

## Summary

This introduces a small, opt-in unit test harness for pure mod logic.

The intent is to establish a lightweight regression-testing path that contributors can run locally without changing the default build or CI behavior. This first slice deliberately stays narrow: it covers shortcut key-name parsing and modifier-key detection, which are pure helpers used by hotkey configuration, rather than trying to model in-game runtime behavior.

What changed:

- adds a `--tests=y` XMake option that includes local tests only when explicitly enabled
- adds a `stfc-mod-tests` doctest target
- splits `Key::Parse`, `Key::IsModifier`, and the key-name lookup table into a small pure source file that can be compiled without Unity/IL2CPP runtime dependencies
- adds regression tests for common shortcut names, alphanumeric keys, function/keypad keys, punctuation aliases, unknown values, and modifier detection

## Current Known Issues

- reviewer requested that the tests use the main file instead of moving the constant into the test file
- the latest GitHub `build-win` check is failing

## Original Local Validation

```bash
xmake f -p windows -m release -y --tests=y
xmake build stfc-mod-tests
xmake run stfc-mod-tests
xmake f -p windows -m release -y --tests=n
xmake build stfc-community-mod
git diff --check
```

This does not add CI enforcement and is not intended to represent full coverage of the mod's in-game behavior. It is a first regression-test foothold for logic that can be exercised outside the game process.


netniV (changes requested):
I do like the idea of having unit tests to check for valid scenario's etc.

Moving the constant from the main file to the test file feels wrong to me.  The test should utilise the main file.

PR #162, opened 2026-05-07
fix(mac): BuffService.ParseBinaryObject signature changed,

a5ehren:
Hook no longer needed because it just pulls out a pointer and calls another function we hook.

### DIFF
--- a/mods/src/patches/parts/sync.cc
+++ b/mods/src/patches/parts/sync.cc
@@ -2171,23 +2171,23 @@ void InstallSyncPatches()
     }
   }
 
+#if __APPLE__
+  // 1.000.49105: BuffService.ParseBinaryObject is a 0x18-byte body immediately before HandleResponseData.
+  // Spud's ARM64 absolute jump is larger than that, so detouring it overwrites the next function entry.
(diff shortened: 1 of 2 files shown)


mods/src/patches/parts/sync.cc
```diff
@@ -2168,11 +2176,17 @@ void InstallSyncPatches()
       !buff_service.isValidHelper()) {
     ErrorMsg::MissingHelper("Services", "BuffService");
   } else {
+#if __APPLE__
+    // 1.000.49105: BuffService.ParseBinaryObject is a 0x18-byte body immediately before HandleResponseData.
+    // Spud's ARM64 absolute jump is larger than that, so detouring it overwrites the next function entry.
+    spdlog::info("Skipping BuffService.ParseBinaryObject hook on macOS");
+#else
```
netniV: Would it not make more sense to put this around the entire buff service since this is the only place it appears to be used?

netniV (changes requested):

mods/src/patches/parts/sync.cc
a5ehren: sure, incoming

netniV (approved):

PR #163, opened 2026-05-07
Add macOS launcher local sync capture

wycats:
Supersedes #152 to support syncing to disk without changing the core of the syncing logic. It keeps the "HTTP-only" architectural decision, but incorporates an HTTP server so the user doesn't have to run their own to sync to disk.

> [!NOTE]
> Because it's "just" an HTTP server, no C++ mod sync/config files are changed (unlike #152)

## Summary

- add a macOS launcher-managed local HTTP sync capture server
- bind capture to `127.0.0.1` on an ephemeral port with a per-session token
- write a managed HTTP sync target into `community_patch_settings.toml` while capture is running
- capture raw POST `/sync` request bodies to `requests.jsonl`
- add launcher controls for Start Capture, Stop Capture, and Open Folder

### DIFF
--- a/macos-launcher/src/ActionView.swift
+++ b/macos-launcher/src/ActionView.swift
@@ -8,6 +8,7 @@ private let logger = Logger(subsystem: "com.stfcmod.startrekpatch", category: "l
 struct ActionView: View, XSollaUpdaterDelegate {
 
   @StateObject var gameUpdater = GameUpdaterViewModel()
+  @StateObject private var syncCaptureServer = SyncCaptureServer()
   @State private var gameVersion: Int = 0
   @State private var gameUpdateAvailable: Bool = false
   @State private var updating: Bool = false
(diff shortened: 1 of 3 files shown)


netniV:
This looks like a nice start.  Maybe this could be extended to include which types of transactions to sync for more control and less traffic... but on paper seems decent.

Assuming @a5ehren can verify this then it could be an interesting addition.  My only concern is getting the project that we had planned done for a unified launcher.

pggpgg:
Been doing something similar with some custom tooling, and I have to say it's been very useful. Would use it if it was built-in to stfc-mod

a5ehren:
I don't see much reason to embed this in the launcher (which we are planning to replace soon). I will ask the Spock's Club team if they will publish a public JSON schema, otherwise I will publish one with the mod. Once you have that it is trivial to vibecode a client for whatever you're doing.

PR #179, opened 2026-05-15
feat: add option to embed custom loading screen image

ChronoXNL:
- Introduced a new option `bg_image` to specify a PNG file for the loading screen background.
- Implemented functionality to regenerate `embedded_loading_image.h` during the build process if a valid image path is provided.
- The generated header file includes the image data and its size for embedding in the application.


xmake.lua
```diff
@@ -1,5 +1,11 @@
 set_project("stfc-community-mod")
 
+option("bg_image")
+    set_showmenu(true)
+    set_description("Path to a PNG to embed as the loading screen background (regenerates embedded_loading_image.h)")
+    set_default("")
```
netniV, line 6: We should default this to the current image path as we already have one.

netniV (changes requested):

PR #183, opened 2026-06-07
Repair off-screen Kirshara queue advancement (fixed by Scopely)

Guffawaffle:
## Retired off-screen queue completion repair

**THIS WAS FIXED BY SCOPELY.** The original off-screen Kir'Shara combat-completion issue no longer needs this workaround. This PR is closed and should not be used for current recovery work.

The separate unavailable-target/course-response race is addressed in [#287 — Faster Queue Recovery](https://github.com/netniV/stfc-mod/pull/287).


Guffawaffle:
Digit finally fixed this natively (for now).

PR #188, opened 2026-06-20
[codex] Harden shortcut config parsing

Guffawaffle:
## What changed

This hardens shortcut config parsing while preserving the mod's established default-binding behavior.

- Missing shortcut entries continue to use built-in defaults.
- `NONE` explicitly unbinds an action.
- Blank, non-string, or fully unparseable configured values fall back to the built-in shortcut and log the reason.
- Valid tokens in a multi-key value remain usable; invalid tokens are reported and ignored.
- Rendered runtime vars include `[shortcuts_source]` so each shortcut can be identified as coming from `default`, `config`, or `alias:<name>`.
- Canonical `set_hotkeys_disabled` / `set_hotkeys_enabled` are honored while old misspelled aliases remain compatible with warnings.

## Why

A malformed shortcut setting should not accidentally remove a default action. Explicit `NONE` remains available for users who intentionally want to unbind a shortcut.

## Validation

- `git diff --check`
- `xmake -y mods`
- AXF release cycle: build, deploy, hash verification, and game relaunch

## Runtime smoke check

Using `zoom_out = "-"` exercised the fallback path: `-` is not a valid shortcut token (use `MINUS` for that key), so the loader logged the invalid token and restored the built-in zoom-out binding.

### DIFF
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -12,6 +12,7 @@
 #include "defaultconfig.h"
 #include <algorithm>
 #include <cstdio>
+#include <initializer_list>
 #include <iostream>
 #include <ranges>
 #include <string>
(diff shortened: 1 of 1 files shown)


netniV:
The invalid items SHOULD default to the base shortcut.  That is by design.  

Don't have an issue with creating a NONE value though to unbind a shortcut if you really want to.

netniV (changes requested):
The invalid items SHOULD default to the base shortcut. That is by design.

Don't have an issue with creating a NONE value though to unbind a shortcut if you really want to.

netniV (approved):

PR #190, opened 2026-06-27
chore: update the prime headers from game version 1.000.50296

a5ehren:


mods/src/patches/parts/sync.cc
```diff
@@ -2139,7 +2139,10 @@ void InstallSyncPatches()
       SPUD_STATIC_DETOUR(ptr, GameServerModelRegistry_ProcessResultInternal);
     }
 
-    ptr = game_server_model_registry.GetMethod("ParseBinaryObjectsHelper");
```
a5ehren, line 2142: I'm still investigating this change, don't merge until I figure it out

a5ehren: ok, reverted. it is actually worse even if technically correct

netniV (approved):

PR #192, opened 2026-07-02
Honour [patches] switches in release builds; make GameServer::Initialise sync hook optional

trcyberoptic:
## What

Two small changes that let users resolve hook collisions with other mods via config instead of rebuilding:

- **config.cc**: read the `[patches]` hook-group switches in release builds too. Previously they were only compiled in under `_MODDBG`; release builds hardcoded every group on. **Defaults are unchanged** — without any `[patches]` entries the behaviour is identical to before (including `resolutionlistfix` staying off).
- **sync.cc**: new switch `[patches] game_version` (default `true`) to skip only the `GameServer::Initialise` detour. That hook solely captures the game version for the `X-PRIME-VERSION` sync header, which has a static fallback (`headers::primeVersion`), so skipping it degrades gracefully — session id and instance id capture are untouched.

## Why

Other BepInEx/Harmony-based mods may detour the same il2cpp methods as this mod. A native detour (this mod) and a managed detour (Il2CppInterop/Harmony) installed on the same method prologue corrupt each other's trampoline and crash the client during login with an `AccessViolationException` in `il2cpp_runtime_invoke`. With these switches, affected users can turn off just the overlapping hook group on this side and keep every other feature (including the rest of sync).

## Testing

- Built via the repo CI workflow (Windows).
- Verified on client 1.000.50475 / Unity 6000.0.59f2 / BepInEx 6.0.0-be.755:
  - Without config entries: log shows the same Patching/Skipping sequence as the official 1.1.2 build (defaults unchanged).
  - With `[patches] uiscalehooks = false` and the game version hook disabled: log shows `x Skipping 1 of 16 (UiScaleHooks)` and `Sync: skipping GameServer::Initialise hook`; a plugin combination that previously crashed at login (double detour) now runs stably, sync uploads keep working.
  - Note: the in-game run used the switch under its initial name `[sync] hook_game_version`; per review it is now `[patches] game_version` — the rename only moves where the flag is read, the gated code path is identical.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -513,6 +513,8 @@ void Config::Load()
       get_config_or_default(config, parsed, "patches", "resolutionlistfix", DCP::resolutionlistfix, write_config);
   this->installSyncPatches =
       get_config_or_default(config, parsed, "patches", "syncpatches", DCP::syncpatches, write_config);
+  this->installGameVersionHook =
+      get_config_or_default(config, parsed, "patches", "game_version", DCP::game_version, write_config);
   this->installObjectTracker =
       get_config_or_default(config, parsed, "patches", "objecttracker", DCP::objecttracker, write_config);
   this->installLoadingScreenBgHooks =
@@ -629,8 +631,6 @@ void Config::Load()
 
   this->sync_debug   = get_config_or_default(config, parsed, "sync", "debug", DCS::debug, write_config);
   this->sync_logging = get_config_or_default(config, parsed, "sync", "logging", DCS::logging, write_config);
-  this->sync_hook_game_version =
-      get_config_or_default(config, parsed, "sync", "hook_game_version", DCS::hook_game_version, write_config);
   this->sync_resolver_cache_ttl =
       get_config_or_default(config, parsed, "sync", "resolver_cache_ttl", DCS::resolver_cache_ttl, write_config);
 
--- a/mods/src/config.cc
+++ b/mods/src/config.cc
@@ -484,7 +484,9 @@ void Config::Load()
     write_log    = false;
   }
 
-#if _MODDBG
+  // Patch switches are honoured in release builds too (previously debug-only):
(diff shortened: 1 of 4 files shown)


netniV:
Thanks for the contribution!  

Reviewing this, I would prefer to give the patching of the game version the same treatment as other patches under [patches] section, renaming it to just `game_version`.  I am a little bit loath to allow selective patching because if people do things that relying on a patch from one side and one from another it could lead to unpredictable results.

The main reason I have them there is that _MODDBG is the flag I run the production code but with developer info and the ability to switch off patches individually to see if they are causing issues.  Not something that general people should be doing.  Also, if a person has the patches flags set and gives someone else a copy of their config, it will also be disabled which may not be the desired result....

Makes me wonder if we should move patches to their own config file because it's a developer specific item?

trcyberoptic:
Moved the switch to `[patches] game_version` in 3a87d34 so it gets the same treatment as the other groups.

On the config-sharing concern: agreed — and as-is this PR actually goes a step further than deliberate sharing. Since `Config::Load()` regenerates a missing `community_patch_settings.toml` from the parsed table, a fresh install now gets the full `[patches]` section written into the auto-generated settings file, effectively advertising the switches as end-user knobs.

A separate file would solve both: read `[patches]` from e.g. `community_patch_patches.toml` if it exists, never auto-create it, never write it back. The main settings file stays clean, shared configs can't silently disable patches, and someone who needs to resolve a hook collision with another mod (a native and a managed detour on the same il2cpp prologue crash at login; the current alternatives are dropping the mod entirely or a custom build) can still create the file deliberately.

Happy to implement that in this PR — just let me know the filename you'd prefer, and whether `_MODDBG` builds should keep reading `[patches]` from the main settings file as they do today.

netniV:
I'll merger as is, then create another PR to move them to their own patch file so they can be turned off without affecting users.

PR #193, opened 2026-07-03
[codex] add mission hud visibility settings

Guffawaffle:
## Summary

Adds experimental per-button Mission HUD visibility controls.

- Configure the buttons in the existing `[ui]` block:
  - `hud_q_trials`
  - `hud_field_training`
  - `hud_outposts`
  - `hud_daily_goals`
  - `hud_missions`
- Each accepts `always`, `auto`, or `never`; all default to `auto`.
- Install the narrow `MissionsHudViewController.UpdateButtons()` hook only when at least one button is configured away from `auto`.
- The hook calls the original first, then applies the configured button GameObject visibility.
- Merged current `dev` and resolved the config conflict, retaining the current release patch-switch behavior.

## Validation

- `git diff --check`
- `xmake -y mods`
- `xmake -y stfc-community-mod`


netniV:
I would prefer not to have a dotted form just for hud items.  I feel these can be in the main block as `hud_xxx`.

netniV (changes requested):
I would prefer not to have a dotted form just for hud items. I feel these can be in the main block as hud_xxx.

mods/src/config.cc
```diff
@@ -716,15 +694,16 @@ void Config::Load()
...
   this->mission_hud_buttons.emplace(
-      "outposts", get_mission_hud_visibility(config, parsed, "outposts", DCMH::outposts, write_config));
+      "outposts", get_mission_hud_visibility(config, parsed, "hud_outposts", DCMH::outposts, write_config));
   this->mission_hud_buttons.emplace(
-      "daily_goals", get_mission_hud_visibility(config, parsed, "daily_goals", DCMH::daily_goals, write_config));
+      "daily_goals", get_mission_hud_visibility(config, parsed, "hud_daily_goals", DCMH::daily_goals, write_config));
   this->mission_hud_buttons.emplace(
-      "missions", get_mission_hud_visibility(config, parsed, "missions", DCMH::missions, write_config));
+      "missions", get_mission_hud_visibility(config, parsed, "hud_missions", DCMH::missions, write_config));
   this->installMissionHudTweaksHooks = this->MissionHudTweaksEnabled();
```
netniV: Since we are now using UI, DCMH is an incorrect prefix.

netniV (changes requested):
Almost there :)

netniV (approved):

PR #194, opened 2026-07-04
Fix macOS zoom range cap

Guffawaffle:
## Summary
- repair the macOS system zoom range during `NavigationZoom.Update`
- keep `_maximum` and `_zoomtotal` aligned with the configured max zoom while in solar-system view
- factor the existing far-ratio and far-clip writes through small helpers

## Root cause
After the recent game update, macOS could enter system view with the game-owned `NavigationZoom` range still capped near the stock max distance, around 2500. The existing macOS path updated far ratios during view setup, but that was not enough to keep the effective `_maximum` range extended during runtime zoom input. Windows was not affected in the same way because it has an additional `SetDepth` hook that macOS cannot safely use.

## Validation
- `git diff --check`
- `xmake -y mods`
- GitHub Actions `Build` on fork branch `fix/mac-zoom-apply-range`
- macOS smoke test confirmed zoom-out works again

### DIFF
--- a/mods/src/patches/parts/zoom.cc
+++ b/mods/src/patches/parts/zoom.cc
@@ -33,9 +33,61 @@ inline void StoreZoom(std::string label, float &zoom, NavigationZoom *_this)
   spdlog::info("Changing {} from {} to {}", label, old_zoom, zoom);
 }
 
-static float          s_expectedScale = 0;
-static void          *s_cachedFR      = nullptr;
-static NavigationZoom *s_navZoom     = nullptr;
+static float           s_expectedScale = 0;
+static void           *s_cachedFR      = nullptr;
+static NavigationZoom *s_navZoom       = nullptr;
+
+static void ApplySystemZoomRange(NavigationZoom *_this, float radius)
+{
+  if (!_this || radius <= 0.0f) {
+    return;
+  }
+
+  auto ratio                     = (Config::Get().zoom / radius);
+  _this->_farRatioSystemNormal   = 0.55f * ratio;
+  _this->_farRatioSystemExtended = ratio;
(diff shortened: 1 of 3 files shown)


mods/src/patches/parts/zoom.cc
```diff
@@ -75,6 +127,10 @@ void NavigationZoom_Update_Hook(auto original, NavigationZoom *_this)
   bool       do_store_zoom    = false;
   auto       config           = &Config::Get();
 
+#if __APPLE__
+  EnsureSystemZoomRange(_this);
+#endif
+
```
netniV: Is this really an apple specific fix, or will PC also be needing this?

mods/src/patches/parts/zoom.cc
```diff
@@ -162,14 +218,19 @@ void NavigationZoom_Update_Hook(auto original, NavigationZoom *_this)
   do_default_zoom = false;
 
   original(_this);
+
+#if __APPLE__
+  EnsureSystemZoomRange(_this);
+#endif
```
netniV: Is this really an apple specific fix, or will PC also be needing this?

netniV (commented):
Two parts of this fix are MacOS specific, but the rest are used regardless.  Will that cause an issue on PC?

mods/src/patches/parts/zoom.cc
Guffawaffle: Good point. The report was Mac-only, but the range correction is not inherently Apple-specific. I removed the Apple guard so the same defensive range refresh applies on PC too. It only runs for solar-system zoom with valid config/radius values, and mirrors the range/far-clip behavior already used by the all-platform depth/view hooks.

PR #197, opened 2026-07-11
Fix loading screen assets across scene transitions

a5ehren:
Root texture and sprite objects with IL2CPP GC handles and recreate them for each login or transition screen generation. 

Read PNG dimensions directly instead of invoking stale Texture accessors. Doing this during the arena load/unload was causing a crash because Unity had GC'd the underlying storage.


netniV (approved):
A lot of changing "Texture" to "Asset" that seems unnecessary but otherwise looks good

PR #203, opened 2026-07-20
Fix macOS ARM64 loading transition crash

Guffawaffle:
## Summary

- preserve register-indirect ARM64 branches such as `BR` and `BLR` when SPUD builds trampolines
- add defensive immediate-operand validation in both relocation paths and bump the vendored SPUD package version
- restore the object-liveness finalize hook and emit a visible macOS ARM64 confirmation log

## Root cause

SPUD classified every ARM64 branch/call operand as an immediate destination. For `BLR x9`, Capstone exposes a register operand; reading the operand union as an immediate turned the x9 register ID (`0xF7`) into a target address. SPUD then sent the instruction through its relative-branch relocator and corrupted the trampoline opcode, producing a `SIGILL` during the loading-screen-to-game-UI transition.

This was not a duplicate or overlapping hook. Register-indirect branches are position-independent and are now copied verbatim, while direct relative branches continue through the existing relocation path.

## Impact

The behavioral change is limited to SPUD's ARM64 relocation classification. Direct `B`, `BL`, `CBZ`, and `TBZ` handling is unchanged; `ADR`/`ADRP` handling is unchanged; and x86/x64 backends are unaffected.

## Validation

- `git diff --check`
- Windows release `xmake` build of `mods`
- SPUD ARM64 sources compiled through the Windows package build
- Capstone operand matrix verified register operands for `BR`/`BLR` and immediate operands for direct branch families
- macOS ARM64 runtime testers confirmed the transition succeeds with object tracking enabled, including arrow-key bindings and bulk auto-claim behavior


mods/src/patches/parts/object_tracker.cc
```diff
@@ -206,6 +206,10 @@ void InstallObjectTrackers()
 
   SPUD_STATIC_DETOUR(il2cpp_unity_liveness_finalize, calc_liveness_hook);
 
+#if defined(__APPLE__) && defined(SPUD_ARCH_ARM64)
```
Guffawaffle, line 209: @netniV this adds a marker to the log file for troubleshooting. Not sure if you mind it or if it should be removed. Personally, I would like to keep it in until we can verify no fall-out.

xmake.lua
```diff
@@ -49,7 +49,7 @@ on_fetch(function(package, opt)
 end)
 package_end()
 
-add_requires("spud v0.2.0-2")
+add_requires("spud v0.2.0-3")
```
Guffawaffle, line 57: I noticed the tashcan's SPUD library 404s and couldn't really find anything that picked up the work. Since there are changes made to that library with this branch, though, it makes sense to me to have an internal version bump. Can re-align if requested.

a5ehren: afaik the entire spud lib is in-tree here

PR #208, opened 2026-07-30
feat: cargo format hooks (configurable decimal precision)

ChronoXNL:
## Summary

- Adds a new opt-in patch (`cargoformathooks`, default `false`) that overrides the number of decimal places shown for abbreviated cargo values in the game UI (e.g. `1.25M` instead of `1M`).
- Hooks `ColourTextLocalizer.SetLocalTextParameters` and temporarily overrides `TextLocalizer.m_significantDecimals` for cargo displays only (identifier `shared_x_of_y_x_coloured`), restoring the original value afterwards via an RAII guard.
- New config settings: `cargoformathooks` (under `[patches]`) and `cargo_significant_decimals` (under `[ui]`, default `2`).

### Safety hardening

- **RAII exception safety** — `FieldRestore` guard restores `m_significantDecimals` even if the original method throws a managed exception that unwinds the stack, preventing permanent localizer corruption.
- **Config clamping** — `cargo_significant_decimals` is clamped to `[0, 6]`; negatives become `0`, values above `6` become `6`. No silent disable sentinel.
- **Fast-path identifier check** — compares `Il2CppString::length` first, then raw UTF-16 chars directly, avoiding a `to_string` allocation on every `ColourTextLocalizer` call (the hook fires for all of them, not just cargo).
- **Compile-time length** — `kCargoIdentifierLen` derived via `sizeof(...) - 1` so it can never drift from the string literal.
- Fields resolved dynamically via the IL2CPP field API — no hardcoded offsets.

#### Test plan

- [x] Builds clean on Windows x64 release (`xmake f -p windows -m release && xmake -y stfc-community-mod`)
- [x] Class hierarchy verified against game dump (`ColourTextLocalizer : TextLocalizer : GraphicLocalizer`)
- [x] `SetLocalTextParameters(bool, object[])` signature matches dump
- [x] Deployed and tested in-game — cargo values display with configured decimal precision
- [ ] Verify with `cargo_significant_decimals = 0` (should show `1M`)
- [ ] Verify with `cargo_significant_decimals = 3` (should show `1.250M`)
- [ ] Verify non-cargo `ColourTextLocalizer` instances are unaffected

Generated with [Devin](https://devin.ai)

### DIFF
--- a/example_community_patch_settings.toml
+++ b/example_community_patch_settings.toml
@@ -229,6 +229,10 @@ loadingscreenhooks = true
 transitionscreenhooks = true
 focussearch = true
 
+# Set to true to override the number of decimal places shown for abbreviated cargo values
+# (e.g. 1.25M instead of 1M). When enabled, set cargo_significant_decimals in the [ui] section.
+cargoformathooks = false
+
 #  <[=========================================================================================================================================]>
 #
 #       *   *          *     *                                      *          ****  *                     *                    *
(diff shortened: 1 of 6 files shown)


ChronoXNL:
Closing — created in error, should target STFC-Statistics/dev.

PR #212, opened 2026-07-31
Fix stale NavigationZoom camera access

wycats:
## Problem

The extended zoom patch retained a `NavigationZoom` instance in `s_navZoom` and reused it from `PlanetViewUtils_CameraZoomedEventHandler_Hook`. That callback can run after a view change, when the cached instance and its scene camera belong to the previous view. Reading `clearFlags` through that stale camera caused an `EXC_BAD_ACCESS`.

## Fix

Remove the cached `NavigationZoom` instance and the callback block that uses it. Camera far-clip configuration already runs from live `NavigationZoom` instances in the view-parameter and depth hooks, so the removed update was redundant.

This keeps camera access tied to the lifetime of the owning view and does not change extended zoom behavior.

### DIFF
--- a/mods/src/patches/parts/zoom.cc
+++ b/mods/src/patches/parts/zoom.cc
@@ -44,7 +44,6 @@ inline void StoreZoom(std::string label, float &zoom, NavigationZoom *_this)
 
 static float           s_expectedScale = 0;
 static void           *s_cachedFR      = nullptr;
-static NavigationZoom *s_navZoom       = nullptr;
 
 static void ApplySystemZoomRange(NavigationZoom *_this, float radius)
 {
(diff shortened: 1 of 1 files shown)


kcid9660:
derp - should've looked for this earlier.  independent confirmation of the same defect anyway.  pulled from a crash minidump on the current client and symbolized :
```
ACCESS_VIOLATION, read of 0x0 faulting frame: 
PlanetViewUtils_CameraZoomedEventHandler_Hook, reading cam->clearFlags
(disasm: `mov eax, [rax]` with rax = 0, clearFlags getter returned null & dereferenced)

game-side call chain:
  NavigationManager.CenterOnFocusPosition
    → NavigationZoom.ChangeToLevel
      → NavigationZoom.ReportZoomChange        (fires the camera-zoomed event)
        → CameraZoomedEventHandler → cam->clearFlags → deref of null
```

`cam` comes from the cached `s_navZoom` after a view change, same stale-instance path this PR removes.  your `EXC_BAD_ACCESS` and my `ACCESS_VIOLATION` <3

my trigger was a zoom / view-change during load and just after a relogin (racing the scene camera), which lines up with your "the callback can run after a view change, when the cached instance and its scene camera belong to the previous view" I think.  I'd independently sent #220, a narrower guard (null-checking `Camera` getters, same as `__get_backgroundColor`) before I found this PR, tho removing the stale cache & redundant block is the cleaner fix.

netniV:
Thanks for the patch, I believe this was fixed in #212 so closing for now.  If you have an updated patch, please feel free to submit it ;)

PR #220, opened 2026-08-08
fix: null-check Camera property getters to prevent a zoom-time crash

kcid9660:
Camera::__get_clearFlags / __get_farClipPlane / __get_nearClipPlane dereferenced GetRaw()'s result without a null check, unlike __get_backgroundColor which guards it.  If the scene camera is stale and a zoom/view-change event fires during load or just after a (re)login the cached NavigationZoom's _sceneCamera resolves to a dead object- the property getter returns null and `*ptr` faults.

Traced from a minidump crash :  bt is in PlanetViewUtils_CameraZoomedEventHandler_Hook reading cam->clearFlags, reached via NavigationZoom.ReportZoomChange on a view change.  So, guard the three getters the same way backgroundColor already does.  The setters use SetRaw (invoke-and-ignore), and seemingly already stale-safe.

### DIFF
--- a/mods/src/prime/Camera.h
+++ b/mods/src/prime/Camera.h
@@ -21,7 +21,8 @@ public:
   float __get_farClipPlane()
   {
     static auto field = get_class_helper().GetProperty("farClipPlane");
-    return *field.GetRaw<float>(this);
+    auto*       ptr   = field.GetRaw<float>(this);
+    return ptr ? *ptr : 0.0f;
   }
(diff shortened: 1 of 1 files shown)


kcid9660:
you probably want #212, ...or both?

kcid9660:
Superseded by #215 (and #212), which remove the stale caches at the root rather than guarding the read — deferring to those. Thanks!

PR #221, opened 2026-08-10
Restore sync compatibility with M93.3.0

lightbull-stfc:
Client version 1.000.51336 (M93.3.0) removed several hooks previously used to synchronize resources and starbase modules. This PR adapts the mod to those client changes and restores the affected sync functionality.

## Changes
* Remove hooks that are no longer present in the game client
* Refactor `sync.cc` around the current client behavior
* Restore resource and starbase module synchronization
* Update protobuf message definitions
* Update the `EntityGroup` enum with newly available values
* Bump the mod version to 1.1.6


Guffawaffle:
Thanks for jumping on this — the protobuf/enum updates and the central `ProcessResultInternal` approach are a really useful fix. I pulled this PR as the base for a client-253 validation pass.

I did find one safety issue before merge: in client 253, `SlotDataContainer.ParseEntitySlotsData` now accepts `EntitySlotsData`, but the detour still declares that argument as `EntityGroup*` and passes it to `HandleEntityGroup`. That can interpret the wrong IL2CPP object layout. Since `ProcessResultInternal` already observes the original `EntityGroup` and covers the full slot snapshot path, I think the safest fix is simply to remove the `ParseEntitySlotsData` detour.

It would also be good to resolve the surviving hooks by exact return/parameter signature rather than method name alone, so future client drift fails closed instead of attaching to a same-name incompatible method.

For the positive news: on a validation branch based on this PR, after removing that slot detour and adding the signature gates, the client booted cleanly and the configured receiver accepted mission, forbidden-tech, officer, module, trait, research, ship, and resource uploads with HTTP 200/204 responses.

Really nice work restoring the actual sync paths here — the concern above is narrow and should be straightforward to address.

lightbull-stfc:
> I did find one safety issue before merge: in client 253, `SlotDataContainer.ParseEntitySlotsData` now accepts `EntitySlotsData`, but the detour still declares that argument as `EntityGroup*` and passes it to `HandleEntityGroup`. That can interpret the wrong IL2CPP object layout. Since `ProcessResultInternal` already observes the original `EntityGroup` and covers the full slot snapshot path, I think the safest fix is simply to remove the `ParseEntitySlotsData` detour.

Good catch! Removed in [135014d](https://github.com/netniV/stfc-mod/pull/221/commits/135014dee8848b06600b523385af0d7acb673de0)

mods/src/version.h
```diff
@@ -3,7 +3,7 @@
 // clang-format off
 #define VERSION_MAJOR               1
 #define VERSION_MINOR               1
-#define VERSION_REVISION            5
+#define VERSION_REVISION            6
```
netniV: Should update the patch version as this isn't a  new release yet.  Odd number releases are now dev builds, even numbers actual releases to make it cleaner for helpers and end users to know which build someone has.  Ultimately, maybe the dev builds include the commit hash but that would need a double commit process (commit src, commit verseion.h  update) with previous hash).

netniV (changes requested):

PR #236, opened 2026-08-20
fix: avoid macOS cache-reset object-tracker crash

Guffawaffle:
## Scope

This PR is intentionally limited to the macOS cache-reset/object-tracker fix. The experimental M94 orphaned keybind-tutorial/`CanUseShortcuts` repair was reverted in `a9a1ae9` after the combined build hung indefinitely on macOS; no crash was observed from that regression. It is no longer present in this branch or its configuration examples.

## Summary

- removes the private `GC_register_finalizer_inner` dependency and the ineffective Unity-liveness cleanup fallback
- tracks objects with weak IL2CPP GC handles instead of raw addresses
- makes `ObjectFinder::GetAll()` return an owned snapshot whose entries hold temporary strong GC-handle leases while callers use them
- prunes handles whose targets are gone while retaining `OnDestroy` as eager cleanup
- updates object-finder callers for the leased result type

## Evidence and root cause

An affected Apple Silicon player consistently hit `EXC_BAD_ACCESS` from `GC_finalize` during **Settings → Clear Cache and Restart**. Disabling `[patches].objecttracker` prevented that crash, isolating the regression to the object-tracker path.

The earlier Apple Silicon fallback skipped native GC finalizer registration and relied on `OnDestroy` plus Unity liveness cleanup. That liveness cleanup was ineffective: IL2CPP's original `LivenessState::Finalize()` clears the mark bit before the mod's subsequent cleanup scan. `ObjectFinder` also exposed raw pointers/spans without synchronized lifetime validation.

The durable fix replaces raw tracked addresses and private finalizers with GC-handle-backed tracking and temporary caller leases.

## Validation

- rollback head `a9a1ae9` has the same Git tree as the object-tracker-only head `9fd9a53`
- Windows debug `mods` build passed for the restored tree
- `git diff --check` passed
- fresh macOS ARM64/x86_64, Windows, packaging, and example-config CI is triggered by the rollback push
- affected-player macOS runtime retest should use the new post-rollback artifact

## Player test plan

1. Install the artifact produced from `a9a1ae9` or later; do not reuse the superseded combined shortcut-gate artifact.
2. Leave `[patches].objecttracker = true`.
3. Run **Clear Cache and Restart**, then repeat the Galactic Anomaly/Academy interaction path that produced the `.ips` crashes.
4. Confirm events, zoom, mod hotkeys, and ordinary cold restarts still work.
5. If a crash recurs, attach the new `.ips` report and surrounding mod log.

## Risk and fallback

The change affects shared object lookup/lifetime behavior, so affected-player validation remains important even though the implementation removes the private finalizer dependency. `[patches].objecttracker = false` remains the emergency fallback.

The orphaned keybind-tutorial/`CanUseShortcuts` issue is now out of scope for this PR and should be investigated separately using hang evidence such as a process sample/spindump, the mod log, and reproduction timing. An `.ips` crash report is not expected unless macOS independently records one.


Guffawaffle:
### Combined macOS test-build checkpoint

Commit `5f47638` is ready for affected-player validation, and all current CI checks are green (macOS ARM64/x86_64, Windows, macOS packaging, and example configs).

Please test this as two independent passes:

1. **Cache-reset/GA stability:** keep `objecttracker = true`, run **Clear Cache and Restart**, then repeat the Galactic Anomaly/Academy path that produced the `.ips` crashes.
2. **Cold-restart Hotkey recovery:** set `[input].repair_orphaned_tutorial_shortcut_gate = true`, cold-restart, and confirm native Mouse Scroll returns after the guarded repair runs. The repair remains `false` by default for everyone else.

A pass/fail result for each item is useful even if the other behaves differently. If either fails, please include the new `.ips` file (if generated) plus the surrounding object-tracker/`[Hotkeys]` log lines so the two regressions stay distinguishable.

mods/src/patches/parts/object_tracker.cc
```diff
@@ -6,183 +6,258 @@
...
+  if (found == tracked_objects_by_class.end()) {
+    return {};
   }
 
-#define GET_CLASS(obj) ((Il2CppClass*)(((size_t)obj) & ~(size_t)1))
-  spdlog::trace("Clearing {}({})", (void*)_this, GET_CLASS(object->klass)->name);
-  remove_from_tracking_all(_this);
-#undef GET_CLASS
+  for (auto record = found->second.rbegin(); record != found->second.rend(); ++record) {
+    auto* target = il2cpp_gchandle_get_target((*record)->weak_handle);
```
a5ehren, line 159: A non-null `il2cpp_gchandle_get_target()` only proves that the managed wrapper is reachable; it does not prove that the Unity native object behind `m_CachedPtr` is still alive. During a reload, the native component can be destroyed while its wrapper remains reachable, and promoting that wrapper to a strong handle can preserve a stale object. `OnDestroy` is useful eager cleanup, but this acquisition path should also validate Unity native liveness (for example via `UnityEngine.Object.op_Implicit`/the equivalent native-alive check) and purge the record before returning a lease.

mods/src/patches/parts/object_tracker.cc
```diff
@@ -6,183 +6,258 @@
...
   auto  on_destroy   = object_class.GetMethod("OnDestroy");
-  if (seen_ctor.find(ctor) == seen_ctor.end()) {
+
+  if (ctor && seen_ctor.emplace(ctor).second) {
     SPUD_STATIC_DETOUR(ctor, track_ctor);
-    seen_ctor.emplace(ctor);
   }
 
-  if (seen_destroy.find(on_destroy) == seen_destroy.end()) {
+  if (on_destroy && seen_destroy.emplace(on_destroy).second) {
```
a5ehren, line 259: This fails open when `.ctor` resolves but `OnDestroy` does not: the object is still tracked, yet native-destruction cleanup is silently absent, and installation later logs an unconditional success message. The current 1.000.51558 metadata does resolve cleanup paths for all 19 tracked types, but ordinary IL2CPP drift could reintroduce stale native wrappers without any diagnostic. Please report per-type hook coverage and either avoid installing constructor tracking without a verified cleanup path, or make the native-liveness validation in the lookup path authoritative.

PR #239, opened 2026-08-21
fix: extend donation slider on macOS

GingerNinja-STFC:
## Scope

This PR extends the existing donation-slider feature to macOS. The Windows implementation is unchanged; only the macOS path and the platform guard on config loading are affected. No other patch behavior is modified.

## Summary

- removes the `#if _WIN32` guard around `extend_donation_slider` / `extend_donation_max` config loading so the settings are read on macOS too
- adds a macOS-only hook on `UnityEngine.UI.Slider::set_maxValue` that replaces a max of `50.0f` or `100.0f` with the configured `extend_donation_max` when the feature is enabled
- keeps the existing `InventoryForPopup::set_MaxItemsToUse` hook Windows-only, since on macOS (M94+) IL2CPP inlines those property methods and they are never called

## Evidence and root cause

On macOS (M94+), the donation-slider cap is no longer enforced through `InventoryForPopup::set_MaxItemsToUse`. IL2CPP inlines the property method, so the existing Windows detour never fires and the slider stays capped at 50. The cap is instead enforced via the Unity `Slider` component's `maxValue` property.

The fix hooks `UnityEngine.UI.Slider::set_maxValue` on macOS and rewrites the max value to the configured limit when the slider's max is 50 or 100 and `extend_donation_slider` is enabled. The Windows path is left untouched.

## Validation

- built the macOS `mods` target (`xmake f -p macosx -a arm64 -m debug --target_minver=13.5 -y && xmake -y mods`)
- `git diff --check` passed
- Windows path is unchanged and guarded by `#if _WIN32`/`#if __APPLE__`, so no Windows behavior regression

## Player test plan

1. Install the macOS artifact built from this branch.
2. Leave `[ui].extend_donation_slider = true` and `extend_donation_max = 80` (defaults).
3. Open the alliance donation popup and confirm the slider extends to the configured max (80) instead of the stock cap (50).
4. Verify other UI sliders that use max 50 or 100 (e.g. DTI faction store encrypted exchange should now also be extended on macOS).

## Risk and fallback

The `set_maxValue` hook is broad: it intercepts **all** Unity sliders with max 50 or 100, not just the donation slider. An unanticipated slider with the same max would also be rewritten.

### DIFF
--- a/mods/src/patches/parts/misc.cc
+++ b/mods/src/patches/parts/misc.cc
@@ -6,9 +6,7 @@
 #include <prime/Hub.h>
 #include <prime/IList.h>
 #include <prime/InventoryForPopup.h>
-#if _WIN32
 #include <prime/InventoryUseRowWidget.h>
-#endif
 #include <prime/ShopSummaryDirector.h>
 
 #include <il2cpp/il2cpp_helper.h>
@@ -23,13 +21,12 @@
 #include <prime/ActionQueueManager.h>
 #include <prime/InterstitialViewController.h>
 
-#if _WIN32
 namespace
 {
 constexpr int64_t kMaximumChestPurchaseMax = 160;
 }
-#endif
 
+#if _WIN32
 void InventoryForPopup_set_MaxItemsToUse(auto original, InventoryForPopup* a1, int64_t a2)
 {
   if (!a1) {
@@ -48,8 +45,8 @@ void InventoryForPopup_set_MaxItemsToUse(auto original, InventoryForPopup* a1, i
 
   original(a1, a2);
 }
+#endif
 
-#if _WIN32
 // IsChestPurchase is populated too late for the existing MaxItemsToUse setter hook, so adjust the tagged context at
 // the row-render seam instead.
 void InventoryUseRowWidget_SetWidgetData(auto original, InventoryUseRowWidget* widget)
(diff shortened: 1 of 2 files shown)


mods/src/patches/parts/misc.cc
```diff
@@ -39,6 +39,20 @@ void InventoryForPopup_set_MaxItemsToUse(auto original, InventoryForPopup* a1, i
   original(a1, a2);
 }
 
+// On macOS (M94+), the InventoryForPopup property methods are inlined by IL2CPP and never
+// called. The donation slider cap is enforced via the Unity Slider component's maxValue
+// property.
+// NOTE: This hooks ALL Unity sliders with max=50 or max=100, not just the donation slider.
```
a5ehren: I don't think we can accept this until we find a way to limit this to alliance donation. Game devs get mad if let people buy 1000 faction items or w/e

netniV: I think I would agree with @a5ehren on this.  We can get this merged but add it's own configuration option and for now the default there would be 160 max.  We can review how this players out once people are actively reviewing.

Thanks for the patches!

GingerNinja-STFC: Well that's unfortunate, I did quite enjoy the side effect 😅 Any action needed on my end? Otherwise feel free to just close this PR.

netniV (approved):

PR #244, opened 2026-08-22
Bugfix/suppress store popups

GingerNinja-STFC:
### Background

The `disable_first_popup` setting was added in PR #37 to suppress the first ad/interstitial popup. It was disabled in PR #139 with the message "temporarily ignore the disable_first_popup option, as it does not work" and hardcoded to `if (false ...)`. The feature has been dead code on `dev` since.

### Why the original approach didn't work

The original hook intercepted `InterstitialViewController.AboutToShow` and called `CloseWhenReady()` **instead of** calling `original()`:

```cpp
if (Config::Get().disable_first_popup && isFirstInterstitial && _this != nullptr) {
    isFirstInterstitial = false;
    _this->CloseWhenReady();   // <-- called WITHOUT original() first
} else {
    original(_this);
}
```

### How this PR fixes it

The hook now **calls `original()` first** to let the popup initialize properly, then immediately closes it:

```cpp
void InterstitialViewController_AboutToShow(auto original, InterstitialViewController* _this)
{
  original(_this);
  if (Config::Get().disable_first_popup && _this != nullptr) {
    spdlog::debug("InterstitialViewController_AboutToShow: suppressing interstitial popup");
    _this->CloseWhenReady();
  }
}
```

- `original()` runs unconditionally → the view controller goes through its full initialization lifecycle, no empty screen, no unready-state crash.
- `CloseWhenReady()` runs **after** initialization → closes the popup immediately after it appears, so the user never sees it.
- The `isFirstInterstitial` one-shot flag is removed → suppression applies to **all** interstitials, not just the first.

### Additional changes

- **Store offer popup suppression** — new hook on `ShopSceneManager.ShowPlcOfferPopup` that drops the call when `disable_first_popup` is enabled. This popup type doesn't require lifecycle initialization, so skipping the call entirely is safe.
- **Consolidated `ShopSceneManager` lookup** — the class was being resolved twice; now resolved once as `static`.
- **Updated example config comments** (en, de, fr, nl) — `disable_first_popup` comment changed from "Prevent the first popup advert" to "Prevent ad/interstitial popups from appearing" to reflect that it now suppresses all such popups, not just the first.

### How to verify

1. With `disable_first_popup = false` (default) — all popups appear as before.
2. Set `disable_first_popup = true`, restart:
   - Ad/interstitial popups should close immediately (look for `InterstitialViewController_AboutToShow: suppressing interstitial popup` in debug logs).
   - Store pack offers should not appear at all (look for `ShopSceneManager_ShowPlcOfferPopup: suppressing PLC offer popup`).

### Risks

- **`CloseWhenReady()` is still called on every interstitial** — while calling it _after_ `original()` avoids the uninitialized-state crash, it has not been runtime-verified against all popup types. The setting is still marked `⚠️ EXPERIMENTAL` in the config and defaults to `false`.
- **`disable_first_popup` semantics broadened** — the setting previously (when it worked) only suppressed the first interstitial; it now suppresses all interstitials and store offers. The config comment has been updated accordingly.

### DIFF
--- a/example_community_patch_settings_de.toml
+++ b/example_community_patch_settings_de.toml
@@ -579,8 +579,11 @@ auto_confirm_ft_upgrade = false
 # Verhindert, dass die Escape-Taste zum Beenden des Spiels auffordert
 disable_escape_exit = true
 
-# ⚠️ EXPERIMENTELL: Verhindert das Erscheinen des ersten Werbe-Pop-ups
+# ⚠️ EXPERIMENTELL: Verhindert das Erscheinen von Werbe-/Interstitial-Pop-ups
 disable_first_popup = false
+
+# ⚠️ EXPERIMENTELL: Verhindert das Erscheinen von Tutorial-/Hinweis-Pop-ups
+disable_tutorials = false
 
 # ⚠️ EXPERIMENTELL: Verhindert die Funktion der standardmäßigen Bewegungstasten
 disable_move_keys = false
--- a/example_community_patch_settings_de.toml
+++ b/example_community_patch_settings_de.toml
@@ -579,8 +579,8 @@ auto_confirm_ft_upgrade = false
 # Verhindert, dass die Escape-Taste zum Beenden des Spiels auffordert
 disable_escape_exit = true
 
-# ⚠️ EXPERIMENTELL: Verhindert das Erscheinen von Werbe-/Interstitial-Pop-ups
-disable_first_popup = false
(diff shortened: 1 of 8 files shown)


janinheuer-tng:
Seems reasonable, is added ✅ 

I just renamed the old config setting since it wasn't doing anything anyway, but if you prefer keeping them in parallel for some reason I can add the other one back in too

PR #253, opened 2026-08-29
Add cross-platform notification audio cues

Guffawaffle:
Adds configurable audio cues for victory, defeat, armada creation, armada victory and armada defeat.

```toml
[audio]
alert_victory = "success"
alert_defeat = "warning"
alert_armada_created = "info"
alert_armada_battle_won = "success"
alert_armada_battle_lost = "warning"
```

Each defaults to `none`, preserving current behavior. Supported cues are `default`, `info`, `success`, `warning`, `alarm`, `arrival`, `soft`, `ping`, and `repair`.

Reuses the existing toast observer. Both platforms use generated in-memory audio: Windows plays through `PlaySoundW`, and macOS through `NSSound`. macOS playback and callback-thread behavior still need runtime verification.

Mac ARM integration smoke confirmed notification audio on the combined build. Intel Mac runtime testing remains unconfirmed; this is supporting integration evidence, not a new standalone artifact qualification.


mods/src/patches/notification_audio.cc
```diff
@@ -0,0 +1,236 @@
...
+    case NotificationSound::Ping:
+      return "ping";
+    case NotificationSound::Repair:
+      return "repair";
+    default:
+      return "none";
+  }
+}
+
+std::optional<NotificationSound> notification_sound_from_name(std::string_view name)
```
Guffawaffle, line 229: No if-chains. this one might be ok since the || comes into play but this is fatter, right, since it has the || ? Let's consider

netniV (approved):

PR #256, opened 2026-08-30
Fix bulk-claim flyout reopening after reward presentation

Guffawaffle:
## Summary

- drive automatic bulk-claim flyout opening from the native shop-list lifecycle instead of inferred navigation payloads
- reopen the enabled selection drawer when returning from reward presentation
- retain one automatic open per visible landing so manual closure is respected for the rest of that visit
- remove the per-frame update detour, 300-frame retry window, hardcoded section allowlist, category/payload conversion, and generic drawer-bind detour
- resolve every required field, property, callback, and lifecycle method before installing the three scroller detours, failing closed on IL2CPP drift

The native `DrawerContext.Enabled` flag is intentionally the eligibility predicate: opted-in users get automatic opening wherever the game exposes an enabled bulk-selection drawer rather than only through enumerated navigation paths.

## Runtime evidence

With `always_skip_reveal_sequence = false`, bulk claimed from Gifts, completed reward presentation, and returned to the shop; the flyout reopened successfully. Manually closing it during the visit left it closed.

Corrected head `4f045c0` was built and deployed with `ax cycle --BuildMode releasedbg`; build and deployed DLL hashes matched at `3782AAA8429D7AD338B5EA774BEB47667D5EE8B767A482DB59EAE3C6AA5C3191`, all required hooks resolved, and boot completed without mod errors. The target-flow smoke above was performed before the final safety-only removal of the generic drawer-bind detour.

A temporary diagnostic-only tab trace caused a client crash during development and was fully removed before either PR commit.

## Review gate

Evidence-first general, hostile, and independent lanes reviewed receipt-bound head `4f045c0` / diff `f200f3719249949228e00b0aad73c1e642a0d00a`. The generic bind hook and partial-install drift risk were corrected in `4f045c0`; full correction re-review found no remaining confirmed defect. The remaining disclosed risk is runtime coverage outside Gifts for the intentionally native-enabled eligibility scope.

## Validation

- `git diff --check`
- Windows releasedbg `mods` build
- Windows release `mods` build
- exact-head releasedbg deploy/hash/clean boot
- GitHub Actions: Windows build, macOS arm64 build, macOS x86_64 build, macOS package, and example-config validation all passed

## Platform note

The documented local macOS command currently fails during XMake option parsing because this checkout does not define `--target_minver`; macOS compile/package validation comes from CI, and macOS runtime behavior remains untested.

### DIFF
--- a/mods/src/patches/parts/open_bulk_claim_gifts.cc
+++ b/mods/src/patches/parts/open_bulk_claim_gifts.cc
@@ -3,6 +3,7 @@
 
 #include <il2cpp/il2cpp_helper.h>
 #include <spud/detour.h>
+#include <spdlog/spdlog.h>
 
 #include <cstddef>
 
@@ -161,6 +162,7 @@ void InstallGiftsBulkClaimHooks()
   auto shop_list_scroller = ShopListScrollerViewController::ClassHelper();
   if (!shop_list_scroller.isValidHelper()) {
     ErrorMsg::MissingHelper("Digit.Prime.Shop", "ShopListScrollerViewController");
+    spdlog::warn("Skipping GiftsBulkClaimHooks: required IL2CPP dependencies are unavailable; no hooks installed.");
     return;
   }
 
@@ -188,6 +190,7 @@ void InstallGiftsBulkClaimHooks()
   }
 
   if (!can_install) {
+    spdlog::warn("Skipping GiftsBulkClaimHooks: required IL2CPP dependencies are unavailable; no hooks installed.");
     return;
   }
 
--- a/mods/src/patches/parts/open_bulk_claim_gifts.cc
+++ b/mods/src/patches/parts/open_bulk_claim_gifts.cc
@@ -100,17 +100,6 @@ void TryAutoOpenDrawer(ShopListScrollerViewController* controller)
   drawer->OpenViaButtonCallback();
 }
 
-void DrawerWidget_OnDidBindContext(auto original, DrawerWidget* self)
-{
-  original(self);
-
-  if (g_active_scroller != nullptr && g_active_scroller->SelectionDrawer() == self) {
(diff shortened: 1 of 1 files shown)


mods/src/patches/parts/open_bulk_claim_gifts.cc
```diff
@@ -154,207 +71,127 @@ class ShopListScrollerViewController
...
 
-  auto section_manager = il2cpp_get_class_helper("Assembly-CSharp", "Digit.Client.Sections", "SectionManager");
-  if (!section_manager.isValidHelper()) {
-    ErrorMsg::MissingHelper("Digit.Client.Sections", "SectionManager");
-  } else if (auto ptr = section_manager.GetMethod("TriggerSectionChange", 5); ptr == nullptr) {
-    ErrorMsg::MissingMethod("SectionManager", "TriggerSectionChange(section,args)");
-  } else {
-    SPUD_STATIC_DETOUR(ptr, SectionManager_TriggerSectionChange);
+  if (!can_install) {
+    return;
```
netniV, line 194: Since one or more items can prevent the overall install, should record that it was skipped.

netniV (changes requested):

mods/src/patches/parts/open_bulk_claim_gifts.cc
Guffawaffle: Added in 812a710. Both dependency-failure exits now log an explicit `Skipping GiftsBulkClaimHooks` summary stating that no hooks were installed, while retaining the individual missing-dependency errors. This also covers the early return when the scroller class itself is unavailable; the successful install path is unchanged.

Validated locally with a Windows release build and a harness exercising the production installer: the success case plus 11 individual and 2 combined dependency failures. Each failure produced exactly one skip summary and installed zero detours.

netniV (approved):

PR #259, opened 2026-08-31
Add All option for disabled audio events

Guffawaffle:
Adds a case-insensitive `All` option to the existing named audio-event filter:

```toml
[audio]
disabled_events = "All"
```

Suppresses playback-start and resume actions while allowing stop, unload and other cleanup actions through. Existing exact-name filtering remains available; `All` takes precedence when combined with names.

Configured mod notification cues remain independent. This applies to named-string Fabric events, not every game audio path; direct event objects and hashed-ID calls are unaffected.

Reuses the existing audio detour. Depends on #253. macOS runtime and native hook compatibility still need verification.


mods/src/config.cc
```diff
@@ -943,14 +962,34 @@ void Config::Load()
...
+  this->alert_victory =
+      get_notification_sound(config, parsed, "alert_victory", DCA::alert_victory, write_config);
+  this->alert_defeat = get_notification_sound(config, parsed, "alert_defeat", DCA::alert_defeat, write_config);
+  this->alert_armada_created =
+      get_notification_sound(config, parsed, "alert_armada_created", DCA::alert_armada_created, write_config);
+  this->alert_armada_battle_won =
+      get_notification_sound(config, parsed, "alert_armada_battle_won", DCA::alert_armada_battle_won, write_config);
+  this->alert_armada_battle_lost =
+      get_notification_sound(config, parsed, "alert_armada_battle_lost", DCA::alert_armada_battle_lost, write_config);
+  if (!this->installToastBannerHooks
```
netniV: If there is a requirement for `installToastBannerHooks`, surely these should only be queried inside that check?

netniV (changes requested):

mods/src/config.cc
Guffawaffle: Good question. Those calls are configuration parsing/validation rather than runtime sound queries, so I kept them unconditional: that preserves normalized `community_patch_runtime.vars` output and invalid-value diagnostics even when the hook is disabled.

I did centralize the dependency instead. A single toast-alert catalog now owns the five ToastState/config/default/member mappings and drives all three uses: parsing, the “any alert configured” warning check, and runtime toast-state lookup. `installToastBannerHooks` remains the sole runtime hook owner; no extra detour is added or auto-enabled.

The owning correction is b39cf0a on prerequisite PR #253 and PR #259 is restacked at 7e75d03. Releasedbg builds, all 11 example configs, three-lane correction review, and Windows runtime testing all passed.

PR #261, opened 2026-08-31
Show mod bindings in native shortcut hints

Guffawaffle:
## Summary

- add a mod shortcut for toggling Scopely's native on-screen keybinding overlay
- replace supported native shortcut badges with the active mod bindings while mod hotkeys are enabled
- mark unsupported native-only rows as unbound because their callbacks are suppressed in mod-hotkey mode
- keep the shipped toggle unbound (`NONE`): no hint-specific hooks, native dependency lookups, or label-cache formatting until a valid toggle binding is configured
- show the first configured alternative in compact badges

The toggle uses Scopely's live `ShowKeybindings` property and preserves its `CanUseShortcuts` gate. When mod hint support is active, disabling mod hotkeys closes an open overlay. Each reveal refreshes the native rows before applying mod labels, and each badge update clears any prior localization override, so startup-disabled, re-enabled, inactive, and reassigned rows cannot retain stale labels. Hint setup validates all required native properties, fields, and methods before installing either hint detour; missing dependencies leave the mod toggle inactive.

## Compact modifier notation

The native badge is only wide enough for a few characters, and STFC's font rendered the platform-native Unicode modifier glyphs as missing-character diamonds. The overlay therefore uses AutoHotkey's established ASCII modifier notation:

| Badge | Modifier |
| --- | --- |
| `^` | Ctrl |
| `!` | Alt / AltGr |
| `+` | Shift |
| `#` | Win / Command |
| `<` / `>` prefix | Left / right modifier |

This keeps common chords readable (`+E`, `!C`, `^+F9`) without truncation. Literal modifier-like keys use unambiguous labels such as `PLS`, and directional keys use `LT`, `RT`, `UP`, and `DN`. A one-line legend is included beside `toggle_shortcut_hints` in every localized example TOML. See [AutoHotkey hotkey modifier symbols](https://www.autohotkey.com/docs/v2/Hotkeys.htm#Symbols).

Compact labels are prepared once after all bindings are parsed, only when the toggle has a valid binding and native dependencies are available. Only the first alternative is formatted; rendering reuses the cached label. Hiding the overlay does not uninstall hooks. Changing the configured binding requires a restart.

## Configuration

```toml
[shortcuts]
toggle_shortcut_hints = "NONE"
```

## Validation at current head

- Head: `1231da6688c67933830e3539efd535d061e0300b`
- Base/merge-base: `0b1c490c967c259186315ceaeb4f5ec2b899ab80`
- Diff hash: `cd68d26c3e8733bb05bb876ef5c7b6a5e4c1912e`
- Receipt: `dd45fd9a432dc6eca4663ca3de40af571c7933238b2476fa3d6c9af6548f5b7d`
- Local Windows x64 releasedbg build, all 11 example TOMLs, and `git diff --check` passed.
- `tests/run-shortcut-hint-cache.ps1` passed against production MapKey/ModifierKey objects, with the Unity-dependent Key seam stubbed. Covers absent/NONE/empty/invalid/modifier-only bindings, valid chords, delayed toggle registration, no parse-time label cache, first-alternative labels, and repeated cache preparation.
- General, hostile, and independent review lanes found no confirmed source defect.
- Current-head GitHub Windows/macOS/config CI is running; no new game deployment or runtime smoke was performed for this correction.

## Historical runtime evidence

Earlier feature heads were manually tested on Windows for show/hide, substituted bindings, compact modifier rendering, and enable/disable transitions. This is not runtime evidence for the current opt-out correction. Earlier Windows corpus inspection found substantive UpdateText/UpdateVisibility bodies (approximately 0x210/0x130 bytes).
## Remaining evidence gates

- cold-start current head with `NONE` to verify no hint detours, then with a valid binding to verify live toggle/refresh behavior

- verify exact-client native extents and live reveal/hide behavior on macOS arm64 and x86_64
- exercise scene recreation, disable/re-enable, and deliberately missing dependency paths
- render a maximal supported chord to determine whether compact badges need an explicit width cap


mods/src/patches/parts/hotkeys.cc
```diff
@@ -52,6 +54,301 @@
...
+    const auto separator = shortcuts.find('-', start);
+    const auto end       = separator == std::string_view::npos ? shortcuts.size() : separator;
+    compact.append(CompactShortcutToken(shortcuts.substr(start, end - start), separator == std::string_view::npos));
+    if (separator == std::string_view::npos) {
+      break;
+    }
+    start = separator + 1;
+  }
+  return compact;
+}
```
netniV: We already loop through the hotkeys at start, surely this should be done as part of that loop?  That way, it can be part of the structure and easily referenced, rather than updated every showing?

netniV (changes requested):

mods/src/patches/parts/hotkeys.cc
Guffawaffle: Good call — the pure binding conversion did not need to run on every reveal. I moved the compact badge generation into `MapKey::Parse()` and added `MapKey::GetShortcutHint()`, so the first configured alternative is now computed once while configuration is loaded. `GetShortcuts()` still retains the complete canonical binding list, including alternatives.

The reveal-time hook remains intentionally: Scopely's `UpdateText` and visibility lifecycle can overwrite or reuse the live `ShortcutKeybindHint` rows, so we still clear/reapply the cached native string when each live row is shown rather than caching managed widget/localizer pointers.

This is in e3f840a. Windows releasedbg builds, clean boot, and repeated F7 show/hide toggles all passed with no shortcut-hint warnings or errors in the log.

mods/src/patches/parts/hotkeys.cc
```diff
@@ -52,6 +54,201 @@
...
+  // Match the native OnShowKeybindingsAction gate while avoiding a fabricated InputAction.CallbackContext.
+  if (!can_use_shortcuts())
+    return false;
+
+  set_show_keybindings(shortcuts_manager, !get_show_keybindings(shortcuts_manager));
+
+  return true;
+}
+
+GameFunction ModFunctionForNativeAction(std::string_view action_name)
```
Guffawaffle: This should be much simpler.

mods/src/patches/mapkey.cc
```diff
@@ -37,19 +137,20 @@ MapKey MapKey::Parse(std::string_view key)
...
-      std::cout << "\n\n----------\nX key:\n" << mapKey << "\n----------\n\n";
+    if (mapKey.Key == KeyCode::X) {
+      std::cout << "\n\n----------\nX key:\n" << mapKey.GetParsedValues() << "\n----------\n\n";
     }
 #endif
   }
 
-  return *mapKey;
+  mapKey.shortcutHint = CompactShortcutForHint(mapKey.Shortcuts);
+  return mapKey;
```
netniV, line 153: The changes here move the allocation from the heap to a local structure, this will potentially cause leaks.

mods/src/defaultconfig.h
```diff
@@ -158,6 +158,7 @@ namespace Shortcuts
   constexpr const char* show_stationexterior  = "SHIFT-G";
   constexpr const char* show_stationinterior  = "SHIFT-H";
   constexpr const char* show_system           = "H";
+  constexpr const char* toggle_shortcut_hints = "NONE";
```
netniV, line 169: Why a string? It's on/off IF it was being used which it is not???

netniV (changes requested):

mods/src/defaultconfig.h
Guffawaffle: This is the key binding for the toggle action, rather than an on/off setting for the overlay itself. For example, `[shortcuts] toggle_shortcut_hints = "F7"` binds the action to F7; the default `"NONE"` leaves it unbound so we do not take another key by default.

It is used: `config.cc` passes it to `parse_config_shortcut(..., GameFunction::ToggleShortcutHints, DCSH::toggle_shortcut_hints)`, and `ScreenManager_Update_Hook` checks `MapKey::IsDown(GameFunction::ToggleShortcutHints)` before calling `ToggleNativeShortcutHints()`. That function toggles Scopely's separate boolean visibility state. The string follows the same convention as the surrounding shortcut bindings.

mods/src/patches/mapkey.cc
Guffawaffle: Let me separate the existing code from the shortcut-hint work, since my earlier explanation was ambiguous.

### What is already in `main`

[`MapKey::Parse()` in upstream main](https://github.com/netniV/stfc-mod/blob/5bada50185f3baf91f84a3622da51e4b798647ee/mods/src/patches/mapkey.cc#L20-L58) returns a **`MapKey` value**, but constructs it this way:

```cpp
auto mapKey = new MapKey();
// Populate mapKey->Modifiers, mapKey->Shortcuts, etc.
return *mapKey;
```

This also exists in the [stable v1.1.6.0 source](https://github.com/netniV/stfc-mod/blob/v1.1.6.0/mods/src/patches/mapkey.cc#L20-L58); it was not introduced by this PR.

### What the ownership bug is

`return *mapKey;` copies the object into the function's return value. It does not return the pointer or transfer ownership of the allocation created by `new`.

The caller receives a valid copy, but the original heap object is never deleted and its pointer is lost. Its owned container storage is therefore left behind too. This occurs when bindings are parsed, not every frame; I am not attributing any observed gameplay memory-growth problem to it.

### How the shortcut-hint work brought this into scope

Following your earlier feedback, [e3f840a](https://github.com/Guffawaffle/stfc-mod/commit/e3f840a0e075f90b2aed213280abbeae6780d0b7) moved compact hint-label generation into `MapKey::Parse()` and added the cached `shortcutHint` string, so opening the overlay no longer repeats that formatting.

That brought us into this existing parsing/ownership path. The later [a6b17c6 optimization](https://github.com/Guffawaffle/stfc-mod/commit/a6b17c6f3f435b77ddae3151836de52a103756df) replaced the raw allocation:

```cpp
MapKey mapKey;
// Populate its owned members.
mapKey.shortcutHint = CompactShortcutForHint(mapKey.Shortcuts);
return mapKey;
```

The new feature did not cause the underlying leak; the ownership cleanup accompanied the work on the parser.

Here, no pointer or reference to the local object escapes. `MapKey` owns its strings and vectors by value. Named return value optimization can construct it directly in the caller's result storage; without that optimization, normal move/copy semantics still preserve ownership and clean up the local object. Correctness does not depend on NRVO being performed. The containers can still allocate on the heap—we are removing the unnecessary separately allocated `MapKey`, not forcing all its data onto the stack.

Technical references: the [C++ return-statement rules](https://eel.is/c++draft/stmt.return#5) specify that the result is initialized before local variables are destroyed; [copy/move elision rules](https://eel.is/c++draft/class.copy.elision#1.1) describe the optional direct construction; the [Core Guidelines' RAII guidance](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rr-raii) explains tying resource ownership to object lifetime.

mods/src/defaultconfig.h
Guffawaffle: Your question also made me realize that leaving the toggle unbound should mean we do not set up the mod's hint support at all. Previously, `"NONE"` left the action unbound, but we still installed the hint-specific hooks and formatted the cached labels.

I've changed that in 1231da6. After configuration is parsed, we check whether the toggle actually has a valid registered binding. If it doesn't, we skip all hint-specific native lookups, both hint detours, and label-cache formatting. The normal hotkey hooks remain in place for the other shortcuts.

So it is still a string because it specifies the toggle key/chord, but `"NONE"` now also opts out of the hint-specific setup. With a valid binding, labels are prepared once at startup and reused when shown—not reformatted on each reveal. Changing the binding requires a restart.

The local Windows build and cache regression tests passed; live-game verification of this latest change is still outstanding.

netniV (approved):

PR #269, opened 2026-09-06
feat: opt-in Windows layout-aware shortcut chords

Guffawaffle:
Adds opt-in Windows keyboard-layout-aware shortcuts. Letters and punctuation follow the active layout, including the Shift needed to type a character—for example, `/` uses Shift+7 on German keyboards.

```toml
[control]
keyboard_layout_mode = "layout"
```

Physical mode remains the default. Restart after changing the setting; subsequent keyboard-layout changes apply live. Displayed bindings retain their configured TOML text.

Mappings refresh through device notifications, without polling. Held keys are suppressed during transitions, and unresolved characters are disabled individually. Characters requiring inferred Ctrl/Alt/AltGr are unsupported.

Windows x64 only. Selecting layout mode on macOS disables printable bindings; use physical mode there. Layout-required modifiers can make existing bindings overlap; see [layout behavior and configuration](https://github.com/Guffawaffle/stfc-mod/blob/feature/layout-aware-letter-shortcuts/docs/keyboard-letter-layout.md).


netniV (commented):
This feels like overkill.  If the unity engine is capable of working both ways, we should be simply following what  the game currently does.  If we should override that, it is surely as simple as updating which mode it's operating in?

Additionally, we are resolving shortcuts during a keyframe in the key functions adding overhead.

Keyboard layouts change by nationality in more ways than just the alphabetic keys.

tests/shortcut_hint_cache.cc
```diff
@@ -56,5 +70,36 @@ int main()
   Check(MapKey::GetShortcutHint(GameFunction::ShowResearch).empty(), "Unbound action acquired a badge");
   MapKey::CacheShortcutHints();
   Check(MapKey::GetShortcutHint(galaxy) == "^G", "Repeated cache preparation changed label");
+
+  // Exercise production action lookup and explicit modifier matching with the
```
Guffawaffle: Hrm, seems sus

tests/shortcut_hint_cache.cc
```diff
@@ -1,23 +1,37 @@
...
+#include "patches/keyboard_layout_mapping.h"
 
 #include <cstdlib>
 #include <iostream>
 
+static std::array<bool, static_cast<int>(KeyCode::Max)> pressed{};
+static keyboard_layout::BindingState layout_bindings;
+static bool layout_enabled = false;
+
 KeyCode Key::Parse(std::string_view key)
```
Guffawaffle, line 18: Still hate if chains

Guffawaffle:
This has been confirmed functional by the requesting user.

PR #273, opened 2026-09-07
fix: preserve system warp destination after zooming out

Guffawaffle:
Fixes primary-action warp after selecting a system and zooming out until its popup disappears. Previously, the dismissed context could submit an invalid destination and leave the ship showing `WARPING 00m 00s` without moving.

For hidden galaxy popups, recover the destination from the selected fleet's matching planned course and submit its explicit position and node ID. Native navigation still handles eligibility, confirmations, cooldown and responses.

Recovery is limited to an unused ordinary system-warp plan whose fleet, origin, path and selected position agree. Invalid or stale data rejects the action. Special travel continues through its normal popup flow.

### DIFF
--- a/mods/src/patches/parts/hotkeys.cc
+++ b/mods/src/patches/parts/hotkeys.cc
@@ -20,6 +20,7 @@
 #include "prime/CanvasController.h"
 #include "prime/ChatManager.h"
 #include "prime/DeploymentManager.h"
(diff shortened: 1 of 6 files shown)


Guffawaffle:
This fails if you switch ships after zooming out. You'll still be the 00m:00s result and it never leaves destination. Converting to draft until I can address that additional repro.

PR #279, opened 2026-09-12
Start config persistence groundwork with checked startup saves

Guffawaffle:
Checks startup configuration output before replacing an existing file, preventing failed or invalid writes from silently replacing the player's configuration.

`Config::Save` validates serialized TOML, checks writing and closing a sibling temporary file, and uses native file replacement. Failures are reported while startup continues with the in-memory configuration. File permissions and recoverable partial-failure files are preserved where supported.

Applies to initial configuration creation and the runtime snapshot. Supersedes #278.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -196,6 +196,16 @@ jobs:
         shell: pwsh
         run: sccache --show-stats
 
+      - name: Test startup config saves
+        shell: pwsh
+        env:
+          PACKAGE_DIR: ${{ steps.xmake_cache_paths.outputs.package_dir }}
+        run: |
+          $header = Get-ChildItem -LiteralPath (Join-Path $env:PACKAGE_DIR 't/toml++') -Recurse -Filter toml.h |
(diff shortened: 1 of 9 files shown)


netniV (commented):
I'm not sure what this for, it seems to add the overhead of saving the configuration file which we never do.

Guffawaffle:
Fair question—the existing writer is adequate for the limited startup use, and this PR's immediate benefit there is modest. Config::Load already creates the config when it is missing and writes the generated runtime snapshot; this doesn't add new save triggers or background work.

I split this out as a small, separately reviewable first step for #280, which supplies persistence for the in-game settings work. Before repeatedly editing the player's actual config, I wanted a checked file-writing foundation: detect write/close failures and stage the new contents before replacing the existing file, rather than truncating it first and potentially leaving a partial file without reporting the failure. The platform-specific handling is there to preserve permissions and deal with replacement failures.

#280 adds the formatting/comment-preserving edits and runtime save coordination separately. I'm happy to trim anything here that isn't pulling its weight; the checked write-and-replace behavior is the part I think is worth establishing before building runtime saves on it. I should have made that split clearer in the PR description.

PR #287, opened 2026-09-16
Faster Queue Recovery

Guffawaffle:
## Summary

When a queued hostile disappears while a course request is outstanding, its failed response can arrive after the game has already removed it. The next target can then wait several seconds for the native watchdog.

This adds opt-in Faster Queue Recovery: an eligible failed response immediately returns control to the native planner to select the next target.

```toml
[control]
faster_queue_recovery = true
```

- Disabled by default; also respects `queue_enabled`.
- Matches the latest observed attempt and queue identity, consumes each recovery once, and preserves native handling when the failed target remains queued by another fleet.
- Uses existing callbacks with bounded weak-reference storage. No timer, frame scan, watchdog hook or per-engagement logging.
- Supports Windows x64 clients with compatible method signatures and queue layouts. Incompatible bindings retain native behavior; other platforms do not install the feature.


Guffawaffle:
This isn't working as well as my previous implementation on my personal branch. Moving back to draft until I figure it out.

PR #292, opened 2026-09-18
Show automatic Warp/Jump choice on the system course button

Guffawaffle:
The system card's SET COURSE label does not show whether automatic confirmation will choose regular warp or instant jump. Update the ordinary course button to WARP or JUMP using the same policy resolver as auto-confirm, including ship-specific overrides. Ask restores the game's native localized wording, and the label refreshes while the card remains open.

Only the action text changes; the timer and click behavior remain native. Toll and wormhole buttons retain their existing wording. Uses the existing screen update dispatcher without adding a detour.

Validation: the identical feature change passed Windows and both macOS architecture builds on the integration stack. Windows player testing confirmed WARP/JUMP changes live on an open card and SET COURSE restoration for Ask, with the timer retained. The standalone Windows release build and diff check passed, and three independent reviews found no actionable defects. Upstream CI is running; macOS runtime is untested.

Mac ARM integration smoke confirmed the system course button displaying WARP for the selected automatic travel mode. This behavior is available on macOS; Intel Mac runtime testing remains unconfirmed.


netniV (changes requested):
There is a conflict that needs resolving, and I'm not entirely sure if some of the methods in warp_action_label.cc are actually elsewhere within the il2cpp code handlers we have.

PR #295, opened 2026-09-19
feat: sync away team assignments

blevouschurverantunt:
## Summary

Adds Away Team data to the sync feature so consumers can tell which officers are currently away and until when. Today `processors::jobs()` skips `JOBTYPE_AWAYASSIGNMENT` and the dispatcher ignores the `AwayAssignments*` entity groups, so a crew-planning tool will happily recommend an officer who is off on a 2-day away mission.

Everything is additive, read-only, reuses the existing `ProcessResultInternal` dispatcher (no new hooks), and is gated like every other data type.

## What changes

- **New sync option** `away_assignments` (`[sync]` and per-target, same semantics/defaults as the other data flags) → new data type `away_assignment`. It is one new row in `SyncOptions`, so `config.cc` needs no changes.
- **`away_assignment` events** from entity group `AwayAssignmentsList` (111, `AwayAssignmentUserListResponse`), only when new/changed:
  ```json
  {"type":"away_assignment","aid":2814298639311718967,"template_id":4186662802,"state":2,
   "officer_ids":[3479870516,440622131,2520801863],"job_uuid":"60f7…264b","duration":43200,"rarity":2}
  ```
  `officer_ids` is emitted in slot order (the proto map iterates in arbitrary order). When an instance disappears from the list a `{"type":"collected_away_assignment","aid":…}` event is sent and its state is pruned, mirroring `completed_job`.
- **`job` events** now include `JOBTYPE_AWAYASSIGNMENT` with `{"aid": …}`, so start time/duration flow through the existing `job` / `completed_job` events (gated by the existing `jobs` option).
- **`officer` events** gain `away_assignment_id` (0 = not away), and a change of that id re-emits the officer (gated by the existing `officer` option). Officer de-dup state is kept local to `officers()` so `forbidden_tech()` is untouched.
- All 10 localized example configs get the new key (first entry of `Subgroup: Data`, alphabetical). The da/de/es/fr/nl/ru comments are best-effort translations in each file's existing style — corrections welcome.

## Verified against the live client (Windows, 1.1.8.0 + this branch, CI-built)

- Group 111 parses as `AwayAssignmentUserListResponse` and re-serialises byte-exact; `officerIds` is slot index → officer id.
- The server sends the list when the Away Teams screen is opened and in the response to collecting a mission (collected instance appears once with state 5, then drops out → `collected_away_assignment`). Lists were always complete, so pruning is safe.
- **Starting a mission** is reflected immediately by the `officer` (`away_assignment_id`) and `job` (`aid`, `start_time`, `duration`) events; the `away_assignment` event for that instance updates the next time the server sends the list. Login alone already yields officer → aid and job(aid) → end time.
- The option is read from `[sync]`, inherited per target, and written to `community_patch_runtime.vars` like the other data flags (it goes through the same `SyncOptions` loop).
- Not handled on purpose: groups 110 (static), 112 and 113 — 113 was never observed during login, opening the screen, or collecting, so nothing unverified is parsed.

Note: `aid` is an int64 above 2^53, emitted as a JSON number like the other ids in the sync stream; JavaScript consumers should be aware of precision.

## Checks

- Fork CI on this exact branch: `build-win`, `build-mac (arm64)`, `build-mac (x86_64)`, `package-mac`, `Validate example configs` — all green.
- `git diff --check` clean, LF only, clang-format applied to the changed lines only.
- No changes to CHANGELOG, README, `version.h`, or `xmake-requires.lock`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


netniV (approved):
Changes look good.  @lightbull-stfc may want to weigh in on whether we need anything more or compatibility for Spocks Club.

mods/src/patches/parts/sync.cc
```diff
@@ -1092,6 +1095,87 @@ static void ship_combat_log_data()
...
+  json event = {{"type", SyncConfig::Type::AwayAssignments},
+                {"aid", instance.id()},
+                {"template_id", instance.awayassignmenttemplateid()},
+                {"state", instance.state()},
+                {"officer_ids", officer_ids},
+                {"job_uuid", instance.jobuuid()},
+                {"duration", instance.parameters().duration()},
+                {"rarity", instance.rarity()}};
+
+  const auto state_value = static_cast<size_t>(std::hash<json>{}(event));
```
lightbull-stfc: Redundant static_cast

mods/src/patches/parts/sync.cc
```diff
@@ -1432,10 +1520,24 @@ static void completed_missions(std::unique_ptr<std::string>&& bytes)
 static void officers(std::unique_ptr<std::string>&& bytes)
 {
   using json = nlohmann::json;
-  using trackers::types::RankLevelShardsState;
 
-  static std::unordered_map<uint64_t, RankLevelShardsState> officer_states;
-  static std::mutex                                         officer_states_mtx;
+  // Local to officers(): like RankLevelShardsState, plus the away assignment instance id so a
```
lightbull-stfc: I would prefer not to have Away Assignments mixed with officers. For users to benefit from this feature addition, they'd currently have to enable away_assignments, officers, and jobs options in the config.

mods/src/patches/parts/sync.cc
```diff
@@ -1455,7 +1558,8 @@ static void officers(std::unique_ptr<std::string>&& bytes)
                                     {"oid", officer.id()},
                                     {"rank", officer.rankindex()},
                                     {"level", officer.level()},
-                                    {"shard_count", officer.shardcount()}});
+                                    {"shard_count", officer.shardcount()},
+                                    {"away_assignment_id", officer.awayassignmentinstanceid()}});
```
lightbull-stfc: This shifts the burden to the sync target to piece back the data together. Slot order is unavailable in this hook and is thus lost. I would prefer to remove it and emit a started assignment through other means.

mods/src/patches/parts/sync.cc
lightbull-stfc: Redundant static_cast

mods/src/patches/parts/sync.cc
```diff
@@ -1446,7 +1548,8 @@ static void officers(std::unique_ptr<std::string>&& bytes)
       std::scoped_lock lk(officer_states_mtx);
 
       for (const auto& officer : response.officers()) {
-        const RankLevelShardsState officer_state{officer.rankindex(), officer.level(), officer.shardcount()};
+        const OfficerState officer_state{officer.rankindex(), officer.level(), officer.shardcount(),
+                                         officer.awayassignmentinstanceid()};
```
lightbull-stfc: I would prefer not to have Away Assignments mixed with officers. For users to benefit from this feature addition, they'd currently have to enable away_assignments, officers, and jobs options in the config.

lightbull-stfc (changes requested):
Let's explore if there's a way to emit updates to away_assignments at a point when an assignment is started (in addition to when the whole list is refreshed) to avoid using officer updates for that.

blevouschurverantunt:
Thanks for the review — all addressed in 6bd1893:

- The `officer` change is dropped entirely; `officers()` is identical to `dev` again.
- The redundant cast is gone.
- Start-time emission: I captured a start with groups 110–113 logged. The server sends group 113 (`AwayAssignmentsInstance`, a bare `AwayAssignmentInstance` that re-serialises byte-exact) when an assignment is started — once with state 1 and the chosen officers, then ~1.5 s later with state 2 and the job uuid. It doesn't appear at login, when opening the Away Teams screen, or on collect, which is why I hadn't seen it before. It now goes through the same helper as the 111 list (shared de-dup state, slot-ordered `officer_ids`), so consumers only need `away_assignments`. With this build, a started assignment reached my sync target as a state-2 `away_assignment` event about a second after the job's `start_time`.

The PR description is updated with the details. Fork CI is green on this exact commit (Windows, both macOS archs, package, config validation); the workflows here still need a maintainer's approval to run.

netniV (changes requested):
Thanks for the update.  Unfortunately, there are now conflicts with this branch.  Could you resolve those so myself and @lightbull-stfc can review once more.

blevouschurverantunt:
Conflicts resolved by merging the current `dev` into the branch (1db8a05). The Haven sync and this change now sit side by side: both options in `SyncOptions`, both dispatcher blocks, no changes to either feature's behaviour. Builds clean locally and on my fork's CI. Ready for another look.

